PR #411, opened 2014-09-18
Use android-maven rather than maven

yaronyg:
The android-maven plugin helps to fix various android specific issues with Maven


hideki:
Current configuration does not have any issue. If we have issue in the future, we will fix this.

PR #479, opened 2014-12-22
Feature/issue 352 get session and checkpoint when retrying replication

hideki:
Add test cases for CBL Java Core #352 

- testCheckSessionAndCheckpointWhenRetryingReplication()
  - When retrying a replication, make sure to get session & checkpoint.
  - This test case must pass with fix of CBL Java Core #299. #299 fix also address this scenario.

- testStopReplicatorWhenRetryingReplicationWithPermanentError()
  - Makes the replicator stop, even if it’s continuous, when it receives a permanent-type error
  - This test case must pass with fix of CBL Java Core #352 on top of #299 fix.

Note: One of test cases depends on fix of CBL Java Core #299. Another one requires CBL Java Core fix of #299 and #352. Otherwise, new test cases causes hang.


tleyden:
Hand merged.

PR #495, opened 2015-01-20
Update Unit Test Cases  to support gzip request and response

hideki:
- Java Core Ticket is #172
- Client sends request with gzip body if SG meets requirements. MockServer needs to parse gzipped request.
- MockServer does send response with gzip.


src/androidTest/java/com/couchbase/lite/replicator/ReplicationTest.java
```diff
@@ -471,7 +475,15 @@ public void changed(Database.ChangeEvent event) {
         // if the server is sync gateway, it is allowable for replicator to use _bulk_get
         RecordedRequest bulkGetRequest = dispatcher.takeRequest(MockHelper.PATH_REGEX_BULK_GET);
         if (bulkGetRequest != null) {
-            String bulkGetBody = bulkGetRequest.getUtf8Body();
+            byte[] body = bulkGetRequest.getBody();
+
+            // for gzip support core java #172
+            // https://github.com/couchbase/couchbase-lite-java-core/issues/172
+            if(bulkGetRequest.getHeader("Content-Encoding")!=null&&bulkGetRequest.getHeader("Content-Encoding").contains("gzip")){
```
tleyden: can you put spaces before and after the "&&"?  it is a bit hard to read.

tleyden:
Instead of `if(request.getHeader("Content-Encoding")!=null&&request.getHeader("Content-Encoding").contains("gzip")){` in several places, can you create a method in MockHelper:

`MockHelper.isGzip(request)` which returns true if gzip?

src/androidTest/java/com/couchbase/lite/mockserver/MockDocumentBulkGet.java
```diff
@@ -72,7 +71,15 @@ public MockResponse generateMockResponse(RecordedRequest request) {
 
         try {
 
-            Map <String, Object> bulkDocsJson = Manager.getObjectMapper().readValue(request.getUtf8Body(), Map.class);
+            byte[] body = request.getBody();
+
+            // for gzip support core java #172
+            // https://github.com/couchbase/couchbase-lite-java-core/issues/172
```
tleyden: I think these comments aren't needed and can be removed, especially if we wrap this in a call like `MockHelper.isGzip(request)`

tleyden:
Are you sure some of the tests cause the replicator use gzip encoding?  I forget the rules that trigger it.

Also are both puller and pusher being exercised with gzip?

hideki:
> Are you sure some of the tests cause the replicator use gzip encoding? I forget the rules that trigger it.

- For request from client to SG, condition is that SG's version > 0.92 and JSON size > 100
- For response from SG to client, it depends on SG. But client needs to set "Accept: gzip" in request header

> Also are both puller and pusher being exercised with gzip?

Yes

hideki:
@tleyden Please re-review updated codes?

PR #500, opened 2015-01-27
Feature/issue 328 slow push

hideki:
Add and modify unit test cases for CBL java core  328 (https://github.com/couchbase/couchbase-lite-java-core/issues/328)

NOTE: New test case for Java Core 328 is created in separated java file, Replication2Test.java instead of ReplicationTest.java. It seems one or some of previously finished test cases make influence to the test result of testExcessiveCheckpointingDuringPushReplication(). I could not identify which one of other test cases cause problem. Instead of identifying the issue, I created new Test class. Also ReplicationTest.java class becomes more than 4K lines.

### DIFF
--- a/src/androidTest/java/com/couchbase/lite/replicator/Replication2Test.java
+++ b/src/androidTest/java/com/couchbase/lite/replicator/Replication2Test.java
@@ -1,0 +1,124 @@
+package com.couchbase.lite.replicator;
+
+import com.couchbase.lite.Document;
+import com.couchbase.lite.LiteTestCase;
+import com.couchbase.lite.mockserver.MockBulkDocs;
+import com.couchbase.lite.mockserver.MockCheckpointPut;
+import com.couchbase.lite.mockserver.MockDispatcher;
+import com.couchbase.lite.mockserver.MockHelper;
+import com.couchbase.lite.mockserver.MockRevsDiff;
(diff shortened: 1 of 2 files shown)


pasin:
@hideki does this PR contain only unit tests () to verify the issue of sending multiple checkpoint requests? Do you plan to add more commits for the fixes?

Regarding "It seems one or some of previously finished test cases make influence to the test result of testExcessiveCheckpointingDuringPushReplication()", does this mean that we have bugs in the unit tests or maybe bugs in the library itself? I think we should fix this.

hideki:
@pasin Actual fix is in Java Core. Ticket is https://github.com/couchbase/couchbase-lite-java-core/issues/328 . It seems PR is already assigned to you.

> Regarding "It seems one or some of previously finished test cases make influence to the test result of testExcessiveCheckpointingDuringPushReplication()", does this mean that we have bugs in the unit tests or maybe bugs in the library itself? I think we should fix this.

I think one of unit tests is not terminate correctly. For example, some of our unit test change static variables in library. Potentially test forget to revert values. Newly added test case always success if it run alone. But it does not work if it run with all other test cases.

tleyden:
> For example, some of our unit test change static variables in library. Potentially test forget to revert values. 

Good catch -- that should definitely get fixed.

PR #504, opened 2015-01-28
Fix #503

### LINKED ISSUE #503: Test06_PullReplication.java and Test07_PushReplication.java are not equivalent with iOS test cases
- Test06_PullReplication.java and Test07_PushReplication.java are not equivalent with iOS test cases
- Test06_PullReplication.java does not pull any documents because remote db and local db are same.

hideki:
Fixed following performance test issues
- iOS test use string as data, but Android test uses array of string as data
- iOS test does not include the attachment, but Android does not.
- Pull replication test never downloaded data because remote db and local db are same.

### DIFF
--- a/src/androidTest/java/com/couchbase/lite/performance2/Test06_PullReplication.java
+++ b/src/androidTest/java/com/couchbase/lite/performance2/Test06_PullReplication.java
@@ -17,40 +17,18 @@
 
 package com.couchbase.lite.performance2;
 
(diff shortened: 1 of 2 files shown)


hideki:
@tleyden If you have time, could you please review the PR. Otherwise, please reassign to @lichenyang2 or @pasin .

tleyden:
> Pull replication test never downloaded data because remote db and local db are same.

Great catch!

PR #521, opened 2015-02-11
Add unit test case for CBL java core 415 https://github.com/couchbase/co...

hideki:
...uchbase-lite-java-core/issues/415

Points of test
- Generate doc with many conflicts with attachment
- atts_since query parameter's value size of /{db}/{docId}?xxx should be 50 or less


pasin:
@hideki It's not mergable. Can you rebase this with the master?

PR #547, opened 2015-03-19
Added comments about Write-Ahead Logging (WAL)

hideki:
- iOS uses Write-Ahead Logging (WAL). But it seem without WAL is little better for Android.


tleyden:
are there any downsides with disabling WAL?

@snej any thoughts on this change?

PR #553, opened 2015-03-25
Fixed CBL Java Core 518

hideki:
https://github.com/couchbase/couchbase-lite-java-core/issues/518

Problem:
From Jenkins log message, CountDownLatch is 0, but it seems task is not done. Maybe because of slow machine. This leads Batcher not to process.

Solution:
Give little bit of time for batcher to finish previous task.

### DIFF
--- a/src/androidTest/java/com/couchbase/lite/support/BatcherTest.java
+++ b/src/androidTest/java/com/couchbase/lite/support/BatcherTest.java
@@ -245,12 +245,9 @@ public class BatcherTest extends LiteTestCase {
      */
     public void testBatcherWaitsForProcessorDelay2() throws Exception {
 
-        long timeBeforeQueue;
-        long timeAfterCallback;
-        long delta;
         boolean success;
 
-        for (int k=0; k<5; k++) {
+        for (int k = 0; k < 5; k++) {
 
             ScheduledExecutorService workExecutor = new ScheduledThreadPoolExecutor(1);
             int inboxCapacity = 5;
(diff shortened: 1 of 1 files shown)


src/androidTest/java/com/couchbase/lite/support/BatcherTest.java
```diff
@@ -285,6 +282,14 @@ public void process(List<String> itemsToProcess) {
...
+            // NOTE: https://github.com/couchbase/couchbase-lite-java-core/issues/518
+            //       CountDownLatch became 0, but task might not be done yet.
+            //       Give little bit of time for slower environment.
+            try {
+                Thread.sleep(500);
+            } catch (Exception ex) {
+            }
+
             // add another object
             batcher.queueObject(new String());
```
pasin: I think this new task added to the batcher got cancelled and it seems like there is a bug in batcher.

pasin: https://github.com/couchbase/couchbase-lite-java-core/blob/b37a40b4786292246aa81c3517b531de0dc83c3b/src/main/java/com/couchbase/lite/support/Batcher.java#L215-L222

When the latch1 became 0 and a new string added to the batcher queue, the inbox size will be more than 0 and the code above will be executed. This could happen in slow environment. I think you can simulate the behavior by adding some sleep right before line 215 in Batcher.java.

If above happens, the new string added to the queue will get cancelled here:
https://github.com/couchbase/couchbase-lite-java-core/blob/b37a40b4786292246aa81c3517b531de0dc83c3b/src/main/java/com/couchbase/lite/support/Batcher.java#L93-L101

pasin: Ignore above. I think my logic is not correct.

pasin: I removed my previous comments to avoid confusion.

1. I think you can simulate the issue by adding some delays at the end of processNow method (L222).

2. As there are some items in the inbox, should it reschedule the processing instead of returning immediately? 
https://github.com/couchbase/couchbase-lite-java-core/blob/b37a40b4786292246aa81c3517b531de0dc83c3b/src/main/java/com/couchbase/lite/support/Batcher.java#L235-L237

hideki: @pasin  Thank you very much for your feedback. Instead of changing BatcherTest, I will try to fix Batcher itself. Thanks!

PR #582, opened 2015-04-16
Ease testing replication with UTF-8 credentials.

sergio91pt:
This way we don't require to encode the username and password manually.

If the username and password is not properly encoded, some tests may fail because of a URL#toURI conversion, that checks for invalid characters.


hideki:
@sergio91pt ,
Thank you very  much for your contribution!

PR #589, opened 2015-04-24
Tests the Query to LiveQuery conversion.

sergio91pt:
Tests that all properties are preserved in the conversion.
More speficially tests that the Query copy constructor copies all fields.

Regression test for couchbase/couchbase-lite-java-core#585.


hideki:
Hi @sergio91pt,  Thank you very much for your contributions!

PR #642, opened 2015-07-27
Fixed #629

### LINKED ISSUE #629: Memory leak/memory management continuous push
We have an app that a couple of times every minute updates a document of around 100kb. No attachments. We then have continuous sync on. After a couple of hours the app runs out of memory, freezes, crashes, can't open more cursors... A bit different behavior depending on app history apparently. Also if there are a lot of changes to be synced, the app can't even be restored after restart. Just opening it can find more than around 1000 changes and then runs out of memory, before able to sync a single record.

Made the following test code that can reproduce the error depending on the sleep time between writes and the servers response time. http://pastebin.com/3mBsENy1

Expected behavior would be to only sync the last revision or at least not trying to load all non synced revisions into memory at the same time, as it looks now, and therefore completely fail with the sync forever until app reinstall.

Client: both couchbase-lite-android.1.0.4 and couchbase-lite-android:1.1.0-532
Server: Both tested with Sync Gateway 1.0.3 and master(bfb402f)

hideki:
- add unit test for newly added utility function.
- update existing unit test for potential test failure.

### DIFF
--- a/src/androidTest/java/com/couchbase/lite/replicator/ReplicationTest.java
+++ b/src/androidTest/java/com/couchbase/lite/replicator/ReplicationTest.java
@@ -3382,11 +3382,13 @@ public class ReplicationTest extends LiteTestCase {
             first = false;
         }
 
+        /* There is a possibility to request indivisual doc. So there is possiblity to fail this test.
         // should not be any requests for individual docs
         for (MockDocumentGet.MockDocument mockDocument : mockDocs) {
             BlockingQueue<RecordedRequest> requestsForDoc = dispatcher.getRequestQueueSnapshot(mockDocument.getDocPathRegex());
             assertTrue(requestsForDoc == null || requestsForDoc.isEmpty());
         }
+        */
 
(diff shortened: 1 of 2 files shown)


pasin:
@hideki I don't understand the root cause of the unit test failure so I don't think I can review this. But I saw some codes got commented out so if those are not applicable anymore we should remove it.

hideki:
https://github.com/couchbase/couchbase-lite-android/commit/a68e8fe8879f545fd95e573bfd0624d8718cc2df => Although Batcher makes tasks processed in a batch. However, because of machine slowness, timing, or other reasons, requests might be contain only one document.  Deleted line of codes could make test failure.

https://github.com/couchbase/couchbase-lite-android/commit/0bfdabdb08761e93e40de17c4c22d4737cf0a186 => In unit test, testContinuousPushReplicationGoesIdleTwice(), local check point may not be sent immediately. Then, deleted lines of codes could make test takes long time to finish.

PR #667, opened 2015-09-22
Integrate with couchbase-lite-java-sqlcipher

pasin:
- Removed libraries/sqlcipher-android
- Linked with couchbase-lite-java-sqlcipher instead of sqlcipher-android
- Reverted SQLiteStorageEngine.open() back to the original that doesn't have encryptionKey parameter.
- Ensured that all methods in AndroidSQLCipherStorageEngine class throws SQLiteException when there is an error,
- Reverted AndroidSQLiteStorageEngineFactory constructor back to the original. There is no need to pass context object to AndroidSQLiteStorageEngineFactory constructor.

Reference: https://github.com/couchbase/couchbase-lite-java/issues/68

### DIFF
--- a/src/main/java/com/couchbase/lite/android/AndroidSQLCipherStorageEngine.java
+++ b/src/main/java/com/couchbase/lite/android/AndroidSQLCipherStorageEngine.java
@@ -52,21 +52,13 @@ public class AndroidSQLCipherStorageEngine implements SQLiteStorageEngine {
     }
 
     @Override
-    public int getVersion() throws SQLException {
-        try {
-            return database.getVersion();
-        } catch (Exception e) {
-            throw new SQLException(e);
-        }
+    public int getVersion() {
+        return database.getVersion();
     }
 
     @Override
-    public void setVersion(int version) throws SQLException {
-        try {
-            database.setVersion(version);
-        } catch (Exception e) {
-            throw new SQLException(e);
-        }
+    public void setVersion(int version) {
+        database.setVersion(version);
     }
 
     @Override
@@ -75,30 +67,18 @@ public class AndroidSQLCipherStorageEngine implements SQLiteStorageEngine {
     }
 
     @Override
-    public void beginTransaction() throws SQLException {
-        try {
-            database.beginTransaction();
-        } catch (Exception e) {
-            throw new SQLException(e);
-        }
+    public void beginTransaction() {
+        database.beginTransaction();
     }
 
     @Override
-    public void endTransaction() throws SQLException {
-        try {
-            database.endTransaction();
-        } catch (Exception e) {
-            throw new SQLException(e);
-        }
+    public void endTransaction() {
+        database.endTransaction();
     }
 
     @Override
-    public void setTransactionSuccessful() throws SQLException {
-        try {
-            database.setTransactionSuccessful();
-        } catch (Exception e) {
-            throw new SQLException(e);
-        }
+    public void setTransactionSuccessful() {
+        database.setTransactionSuccessful();
     }
 
     @Override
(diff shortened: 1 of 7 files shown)


src/main/java/com/couchbase/lite/android/AndroidSQLiteStorageEngine.java
```diff
@@ -32,12 +32,12 @@
     private SQLiteDatabase database;
     private AndroidContext context;
 
-    public AndroidSQLiteStorageEngine(AndroidContext context) {
+    public AndroidSQLiteStorageEngine() {
         this.context = context;
```
hideki: Line 36 - Could be removed.

src/main/java/com/couchbase/lite/android/AndroidSQLiteStorageEngine.java
```diff
@@ -32,12 +32,12 @@
     private SQLiteDatabase database;
     private AndroidContext context;
```
hideki: context is never used. this also removed.

src/main/java/com/couchbase/lite/android/AndroidSQLCipherStorageEngine.java
```diff
@@ -22,89 +22,94 @@
...
             if (database != null)
                 database.close();
             throw new SQLException(e);
         }
-
         return database.isOpen();
     }
 
     @Override
     public int getVersion() {
```
hideki: This is general comments for all methods that are declared in this class.

As all method could throw SQLException, but method signature is not defined with throws SQLException. Is it correct?

hideki:
From https://github.com/couchbaselabs/couchbase-lite-java-sqlcipher
4. The couchbase-lite-java-sqlcipher is bundled natively with REVID and JSON collator used by couchbase-lite-android and couchbase-lite-java.

Should we remove precompiled collators from couchbase-lite-android?

hideki:
https://github.com/couchbaselabs/couchbase-lite-java-sqlcipher
in libs directory, there are precompiled libraries. Could you please explain which libraries should be compiled at build time and which one is not.

Build team is set up NDK on build server (Jenkins). All native libraries should be build on build server.

pasin:
No, we can't remove precompiled collators from couchbase-lite-android. This is for SQLite database from Android SDK.

src/main/java/com/couchbase/lite/android/AndroidSQLiteStorageEngine.java
pasin: Will fix this.

src/main/java/com/couchbase/lite/android/AndroidSQLiteStorageEngine.java
pasin: Will fix this.

src/main/java/com/couchbase/lite/android/AndroidSQLCipherStorageEngine.java
pasin: First SQLException is a RuntimeException so there is no needs to declared. I don't know why SQLiteStorageEngine interface declared SQLException in only a few methods - I think there shouldn't be any thrown declared but I don't want to change that as it will affect current implementation.

For example, there is no exceptions declared for http://docs.oracle.com/javase/7/docs/api/java/lang/Thread.html#interrupt() but it could thrown a few runtime exceptions.

pasin: The same for http://developer.android.com/reference/android/database/sqlite/SQLiteDatabase.html. Several methods can thrown SQLException (runtime as well) without declaring.

pasin: My above comment is wrong. I think it's a good practice (even though no requirements) to declare even runtime exception.

pasin: I'm fixing it now.

pasin:
@hideki I made changes according to your comment and commit. Please review.

pasin:
@hideki I'm revising the throws statement. Please don't review my previous fix yet.

pasin:
I'm done revise catching and throwing the SQLException. It's ready to be reviewed.

PR #670, opened 2015-09-24
Implement insertOrThrow of SQLiteStorageEngine

pasin:
- Implemented insertOrThrow method of the SQLiteStorageEngine interface
- On AndroidSQLCipherStorageEngine, instead of caching a generic exception, caching com.couchbase.lite.database.SQLException instead.


hideki:
Can you implement `replaceOrThrow()` too?

pasin:
Are you using replace method?

pasin:
Currently we don't have replace method in the interface so I assume that there is no codes using replace feature. Do we need replace?

hideki:
OK. Thanks!

PR #677, opened 2015-10-02
Support SQLite database and attachment rekey

pasin:
- Added DatabaseEncryptionTest's testEncryptedAttachments  and rekeyTest
- Added getTempDir to the AndroidContext class.

https://github.com/couchbase/couchbase-lite-java-core/issues/719
https://github.com/couchbase/couchbase-lite-java-core/issues/720

### DIFF
--- a/src/androidTest/java/com/couchbase/lite/DatabaseEncryptionTest.java
+++ b/src/androidTest/java/com/couchbase/lite/DatabaseEncryptionTest.java
@@ -32,25 +32,21 @@ import java.util.Map;
 public class DatabaseEncryptionTest extends LiteTestCaseWithDB {
     private static final String TEST_DIR = "encryption";
     private static final String NULL_PASSWORD = null;
-    private Manager currentManager;
+    private Manager cryptoManager;
 
     @Override
     protected void setUp() throws Exception {
         super.setUp();
         ManagerOptions options = new ManagerOptions();
         options.setEnableStorageEncryption(true);
-        currentManager = new Manager(getTestContext(TEST_DIR, true), options);
+        cryptoManager = new Manager(getTestContext(TEST_DIR, true), options);
     }
 
     @Override
     protected void tearDown() throws Exception {
         super.tearDown();
-        if (currentManager != null)
-            currentManager.close();
-    }
-
-    private Manager getEncryptionTestManager() {
-        return currentManager;
+        if (cryptoManager != null)
+            cryptoManager.close();
     }
 
     public void testEncryptionFailsGracefully() throws Exception {
@@ -78,8 +74,6 @@ public class DatabaseEncryptionTest extends LiteTestCaseWithDB {
     public void testUnEncryptedDB() throws Exception {
         if (!isSQLiteDB())
             return;
-
-        Manager cryptoManager = getEncryptionTestManager();
 
         // Create unencrypted DB:
         Database seekrit = cryptoManager.getDatabase("seekrit");
@@ -110,8 +104,6 @@ public class DatabaseEncryptionTest extends LiteTestCaseWithDB {
     public void testEncryptedDB() throws Exception {
         if (!isSQLiteDB())
             return;
-
-        Manager cryptoManager = getEncryptionTestManager();
 
         // Create encrypted DB:
         cryptoManager.registerEncryptionKey("123456", "seekrit");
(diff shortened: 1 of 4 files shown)


src/androidTest/java/com/couchbase/lite/DatabaseEncryptionTest.java
```diff
@@ -230,4 +234,67 @@ public boolean update(UnsavedRevision rev) {
         Assert.assertNotNull(seekrit);
         Assert.assertEquals(1, seekrit.getDocumentCount());
     }
+
+    public void testEncryptedAttachments() throws Exception {
+        if (!isSQLiteDB())
+            return;
+
+        Manager cryptoManager = getEncryptionTestManager();
```
hideki: `getEncryptionTestManager()` is just return `currentManager`. Why does the method need special naming?

src/androidTest/java/com/couchbase/lite/DatabaseEncryptionTest.java
```diff
@@ -230,4 +234,67 @@ public boolean update(UnsavedRevision rev) {
...
+
+        // Check the document and its attachment:
+        SavedRevision savedRev = seekrit.getDocument("att").getCurrentRevision();
+        Assert.assertNotNull(savedRev);
+
+        Attachment att = savedRev.getAttachment("att.txt");
+        Assert.assertNotNull(att);
+        byte[] body = "This is a test attachment!".getBytes();
+        byte[] rawAtt = TextUtils.read(att.getContent());
+        Assert.assertTrue(Arrays.equals(rawAtt, body));
```
hideki: Does not need to compare between rawAtt which was encrypted with previous key and rawAtt which is encrtypted with current key?

src/androidTest/java/com/couchbase/lite/DatabaseEncryptionTest.java
pasin: I intended to give it the named so that we know that we are getting a manager with an encryption. Nothing more than this.

src/androidTest/java/com/couchbase/lite/DatabaseEncryptionTest.java
pasin: That is the same thing here as we explicitly compare with the "This is a test attachment!" text.

pasin:
I removed the method and change currentManager to cryptoManager and the changed was pushed.

src/androidTest/java/com/couchbase/lite/DatabaseEncryptionTest.java
hideki: This is test for re-key feature. So, updated rawAtt should be different from original, "This is a test attachment!", and also different from previous rawAtt which was encrypted by previous key. So, I recommend to add one additional check, `Assert.assertTrue(Arrays.equals(rawAtt, prvRawAtt));`

pasin: No, the rawAtt is the same as previous-raw-att. The rawAtt is the byte array of the attachment, which is the `decrypted` one not the `encrypted` one.

pasin: Let me clarify more with the code:

// Here is the text content of the attachment converted to raw bytes:
byte[] body = "This is a test attachment!".getBytes();

// Here is the decrypted raw bytes read from the attachment.
byte[] rawAtt = TextUtils.read(att.getContent());

// There is an assertion to check if it's the same raw bytes
Assert.assertTrue(Arrays.equals(rawAtt, body));

// Here is your recommendation which does the same thing as above because they all are decrypted raw bytes:
Assert.assertTrue(Arrays.equals(rawAtt, prvRawAtt));

pasin: Let me know if I misunderstood your point.

hideki: Sorry, I misunderstood the code. I thought it compares encrypted byte arrays.

pasin: No problems. I think the variable name of rawXXX might be confusing a bit.

PR #679, opened 2015-10-09
Refactor performance tests

pasin:
- Created a new base class, "PerformanceTestCase" and new config file, "perftest.properties".
- Removed performance test config out of test.properties.
- Removed performance/Test16_ParallelPushReplication as the test is too complex to run so I don't think we have ever run it.
- Removed all tests in performance2 package and LitePerfTestCase.java
- Ported performance2/Test29_AllDocQuery and performance.
- No port performance2/Test28_KeySizes, Test30_LiveQuery, and Test31_CompactDB as those tests do not make sense or not complete.


hideki:
Hi @pasin,

If you have time, could you please review document size and attachment size?

- Document: Java char is 2 bytes. When it is serialized to JSON, it will be encoded into UTF-8 and it is 1 byte if char is ASCII range character. And UTF-8 json is stored into a database.
- Attachment: Java char is 2 bytes. If it simply wrote to File, it is still 2 byte without encoding to UTF-8 or ASCII.

Currently doc size and attachment size are not clearly defined. 

Thanks,
Hideki

hideki:
This PR is large PR. Basically it looks fine. But I could not check details. So we will fix the issue whenever we find the error in tests.

PR #691, opened 2015-11-06
Replace JNI-based JSON collation with ICU4C

hideki:
Original Ticket: https://github.com/couchbase/couchbase-lite-java-core/issues/801

- JSONCollator used to call java.text.Collator. Instead of java.text.Collator, use ICU4C collator.

Note for reviewer:
- All .h header files are just copied from ICU4C code base for compilation.
- All *.a static libraries are compiled from https://github.com/couchbaselabs/icu4c-android


src/main/java/com/couchbase/lite/android/AndroidSQLiteStorageEngine.java
```diff
@@ -174,4 +184,55 @@ public String toString() {
...
     }
+
+    private static synchronized void loadICU4C(android.content.Context context, File workingDir) {
+        boolean systemICUFileExists = new File("/system/usr/icu/icudt53l.dat").exists();
+        String icuRootPath = systemICUFileExists ? "/system/usr" : workingDir.getAbsolutePath();
+        SQLiteJsonCollator.setICURoot(icuRootPath);
+        if(!systemICUFileExists){
+            loadICUData(context, workingDir);
+        }
+        SQLiteJsonCollator.setLocale(Locale.getDefault().toString());
```
pasin: I have seen that in unit test before calling setLocale(), releaseICU() is called. Should it be the same thing here given that the database can be closed and reopened?

I'm curious that on an android device, if a user change the locale while he is using the app, how can the locale in the collator be updated?

src/main/java/com/couchbase/lite/android/AndroidSQLiteStorageEngine.java
```diff
@@ -174,4 +184,55 @@ public String toString() {
...
+                    out.write(buf, 0, len);
+                }
+            }
+        }
+        catch (Exception ex) {
+            Log.e(TAG, "Error copying icu dat file", ex);
+            if(icuDataFile.exists()){
+                icuDataFile.delete();
+            }
+            throw new RuntimeException(ex);
```
pasin: I think we need to handle this properly in open() method.

src/main/java/com/couchbase/lite/android/AndroidSQLiteStorageEngine.java
```diff
@@ -174,4 +184,55 @@ public String toString() {
...
+                if(in != null){
+                    in.close();
+                }
+                if(out != null){
+                    out.flush();
+                    out.close();
+                }
+            } catch (IOException ioe){
+                Log.e(TAG, "Error in closing streams IO streams after expanding ICU dat file", ioe);
+                throw new RuntimeException(ioe);
```
pasin: Same here.

src/main/jni/JsonCollator/com_couchbase_lite_android_SQLiteJsonCollator.cpp
```diff
@@ -632,3 +574,29 @@ JNIEXPORT jint JNICALL Java_com_couchbase_lite_android_SQLiteJsonCollator_testDi
...
+{
+    char const * ICURootPath = env->GetStringUTFChars(ICURoot, NULL);
+    setenv("CBL_ICU_PREFIX", ICURootPath, 1);
+    env->ReleaseStringUTFChars(ICURoot, ICURootPath);
+}
+
+JNIEXPORT void JNICALL Java_com_couchbase_lite_android_SQLiteJsonCollator_setLocale
+    (JNIEnv *env, jclass clazz, jstring jlocale)
+{
+    char const * localeStr = env->GetStringUTFChars(jlocale, NULL);
```
pasin: Shall we also check NULL on jlocale or it will never be NULL?

src/main/jni/JsonCollator/com_couchbase_lite_android_SQLiteJsonCollator.cpp
```diff
@@ -303,60 +283,22 @@ static int compareStringsUnicodeFast(const char** in1, const char** in2) {
...
-    env->DeleteLocalRef(str1);
-    env->DeleteLocalRef(str2);
+
+    // ICU4C - Collator
+    if(coll == NULL){
+        UErrorCode status = U_ZERO_ERROR; 
+        coll = Collator::createInstance(locale, status);
+        if(!U_SUCCESS(status)) {
+            LOGE("Failed to create Collator instance: status=%d", status);
+            return -3;
```
pasin: What will happen in in CollateJSON() when -3 is returned? It's really not clear from reading CollateJSON() code about what will happen.

src/main/java/com/couchbase/lite/android/AndroidSQLiteStorageEngine.java
hideki: > I have seen that in unit test before calling setLocale(), releaseICU() is called. Should it be the same thing here given that the database can be closed and reopened?
`releaseICU()` should be called from `Object.finalize()`. I forgot to port from v1.1.0 version of implementation. Thank you very much for catching this.

> I'm curious that on an android device, if a user change the locale while he is using the app, how can the locale in the collator be updated?

Ideally, if user changed Locale, it should reindex. Current code does not care this scenario.

src/main/jni/JsonCollator/com_couchbase_lite_android_SQLiteJsonCollator.cpp
hideki: As you suspect, -3 makes nothing. sort orders list as it is.
This method can not handle error. So I will change to 0 as same as kIllegal case.

hideki:
I updated my PR except README. Could you please review PR again?

hideki:
About ICU4C library, I updated REAME.md .

hideki:
@pasin,
Please review the code again?

PR #705, opened 2015-11-24
Fixed #685 compact() does not work as expected.

### LINKED ISSUE #685: Compaction with any revision depth doesn't seem to do anything
Hey there guys,

Our team has a high priority on working with compaction of our databases on our mobile devices, and these last few days have seen a lot of tests on our part. Part of these tests were manually manipulating a single docs's attachments from a *server side implementation* to test how/if the compact() method would work, and it seems that through a manual database manipulation from a *server side implementation* of a doc's attachments, it did (i.e., the removal of one of the sha1's seemed to prompt the remove of the attachment .blob from disk).

However, now that we are attempting to automate this process, the attachments for old revisions do not seem to be removed from disk during compaction. In fact, it seems that compaction of the database does not actually yield any results. This may be due to some bug on our part, but we're not entirely sure.

We have used `database.setMaxRevTreeDepth(X);` to all sorts of values, but no matter the value (0, 1, 2, 5, 10, etc.), the end results are the same - the attachments that were on OLD revisions of documents stick around, and the database size remains unchanged.

Doing a little bit of step-through-debugging, I found the actual lines that seem to be responsible for pruning the DB:
`Database.java`
`pruneRevsToMaxDepth`
`lines : 5191-5196`

The toPrune hashMap that results for our current database during testing ends up at 3091 keys, and during the while loop, `database.delete("revs", "doc_id=? AND revid < ? AND current=0", deleteArgs);` never returns a non-zero value. Not exactly sure if that's a helpful symptom, but it was noticed.

Unfortunately, the next method, `garbageCollectAttachments()` then seems to have a File[] that contains every single attachment in the '_dbname_ attachments' directory, and doesn't delete any of them.

It's very possible I have a misunderstanding with how compaction affects a local device, insofar as modifying a document on a device to change the revision, setting a max depth, and then compact()'ing will not actually delete old revisions for docs without intervention from server. But, it's also possible there's a super subtle bug somewhere in the pruning code. In either scenario, I would absolutely appreciate some help and/or understanding on this. At the very least, it will help us to understand the toolset better!

A great many thanks for your time on this, again!

hideki:
This PR is to update unit test for `compact()`

- Added testRevPruning() into ApiTest.java for test `Database.compact()`
- Renamed AttachmentsTest.java to DatabaseAttachmentTest.java for similar naming with iOS.
- Added  `void test13_GarbageCollectAttachments()`  into DatabaseAttachmentTest.java
- update Database.java for testing `Database.compact()`

### DIFF
--- a/src/androidTest/java/com/couchbase/lite/DatabaseAttachmentTest.java
+++ b/src/androidTest/java/com/couchbase/lite/DatabaseAttachmentTest.java
@@ -956,7 +956,7 @@ public class DatabaseAttachmentTest extends LiteTestCaseWithDB {
         assertTrue(Arrays.equals(bytes, receivedBytes));
     }
 
-    public RevisionInternal putDocWithAttachment(String docID, String attachmentText, boolean compress) throws CouchbaseLiteException {
+    private RevisionInternal putDocWithAttachment(String docID, String attachmentText, boolean compress) throws CouchbaseLiteException {
         byte[] attachmentData = attachmentText.getBytes();
         String encoding = null;
         int length = 0;
--- a/src/androidTest/java/com/couchbase/lite/DatabaseAttachmentTest.java
+++ b/src/androidTest/java/com/couchbase/lite/DatabaseAttachmentTest.java
@@ -36,6 +36,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.net.URL;
 import java.nio.charset.Charset;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -44,7 +45,10 @@ import java.util.Map;
 import java.util.Set;
 import java.util.zip.GZIPOutputStream;
 
-public class AttachmentsTest extends LiteTestCaseWithDB {
+/**
+ * Tests ported from DatabaseAttachment_Tests.m
+ */
+public class DatabaseAttachmentTest extends LiteTestCaseWithDB {
 
     public static final String TAG = "Attachments";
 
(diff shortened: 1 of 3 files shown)


src/androidTest/java/com/couchbase/lite/DatabaseAttachmentTest.java
```diff
@@ -168,6 +173,28 @@ public void testAttachments() throws Exception {
     }
 
     /**
+     * - (void) test13_GarbageCollectAttachments
+     */
+    public void test13_GarbageCollectAttachments() throws CouchbaseLiteException{
```
pasin: To be consistent, we shouldn't prefix with test13 or we can prefix all test methods here with testxx.

src/androidTest/java/com/couchbase/lite/DatabaseAttachmentTest.java
```diff
@@ -931,7 +956,7 @@ public void testGzippedAttachmentByBase64() throws Exception {
         assertTrue(Arrays.equals(bytes, receivedBytes));
     }
 
-    private RevisionInternal putDocWithAttachment(String docID, String attachmentText, boolean compress) throws CouchbaseLiteException {
+    public RevisionInternal putDocWithAttachment(String docID, String attachmentText, boolean compress) throws CouchbaseLiteException {
```
pasin: Why public?

hideki: Initially I tried to use this method from other test class. As I did not do so, It should be stay `private`.

src/androidTest/java/com/couchbase/lite/DatabaseAttachmentTest.java
hideki: In many times, I would like to compare the test code with iOS code. Without same name, it is harder to find test. Personally I'd like to use same name.

PR #711, opened 2015-12-01
With first device or emulator, stop event is fired before mock server…

hideki:
… finished to process request from client. To avoid threading issue, give 1sec to wait for mock server finishes to process all requests.

### DIFF
--- a/src/androidTest/java/com/couchbase/lite/replicator/ReplicationTest.java
+++ b/src/androidTest/java/com/couchbase/lite/replicator/ReplicationTest.java
@@ -1401,6 +1401,9 @@ public class ReplicationTest extends LiteTestCaseWithDB {
             runReplication(pullReplication);
             assertNull(pullReplication.getLastError());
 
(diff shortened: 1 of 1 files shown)


src/androidTest/java/com/couchbase/lite/replicator/ReplicationTest.java
```diff
@@ -1401,6 +1401,9 @@ public void testRemoteConflictResolution() throws Exception {
             runReplication(pullReplication);
             assertNull(pullReplication.getLastError());
 
+            // With the faster device, stop event is fired before mock server receive request.
+            try { Thread.sleep(1000); } catch (Exception ex) { }
```
pasin: Doesn't the replicator need to wait for responses from the Mock Server after sending any requests? Could you explain a bit more?

hideki:
Error could be in MockServer. I will close this PR and will investigate MockServer further.

PR #752, opened 2016-01-12
Fixed Java Core 785 - Parse new compound Server header format

hideki:
- Actually client does not change the code to adapt new Server format.
- This is for adding test


hideki:
merged from master to address conflict.

PR #764, opened 2016-01-27
Running updateIndex twice between creating revs 2 and 3 to verify #952

alexlehn:
Adding a new updateIndex call between saving the conflict versions and the saving the resolutions.

Test Case verifying issue https://github.com/couchbase/couchbase-lite-java-core/issues/952


hideki:
@alexlehn Thank you very much for updating test case.

PR #819, opened 2016-03-26
Fixed java-core 1163 - Replicator sends duplicated Replicator state.

hideki:
Reason duplicated replicator states are notified by `changed()` is because most of unit tests check replicator state which could be already changed to next state. So the replicator enters STOPPING state, then ChangeEvent is fired, next the replicator enters STOPPED state, then ChangeEvent is fired again. When listener receives notification for STOPPING, the replicator state already changed to STOPPED state. So it seems like STOPPED state are fired twice. But actually STOPPING state and STOPPED states are fired. Instead of checking Replicator.status, needs to check transition destination.

### DIFF
--- a/src/androidTest/java/com/couchbase/lite/DatabaseTest.java
+++ b/src/androidTest/java/com/couchbase/lite/DatabaseTest.java
@@ -191,7 +191,7 @@ public class DatabaseTest extends LiteTestCaseWithDB {
             assertEquals(0, database.getActiveReplications().size());
 
             final CountDownLatch replicationRunning = new CountDownLatch(1);
-            replication.addChangeListener(new ReplicationActiveObserver(replicationRunning));
+            replication.addChangeListener(new ReplicationRunningObserver(replicationRunning));
 
             replication.start();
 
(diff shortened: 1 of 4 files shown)


pasin:
What does this mean to users when listening to replication change events? Do the users need to check the state transition instead of replication status to get the accurate info?

hideki:
> Do the users need to check the state transition instead of replication status to get the accurate info?

Currently Yes. `Replication.ChangeEvent.getSource()` returns instance of actual `Replication`, and `ReplicationStatus getStatus()` returns state from `ReplicationInternal`. 

We need to set `ReplicationStatus` to `ChangeEvent`. However this requires API change.

pasin:
I don't know if I understand the issue correctly. If Replication.status is not correct, we should fix that. I will merge the pull request.

hideki:
I will create ticket for Replication.status issue.

Issue is State of replication could be changed to next state when `changed()` callback was called because `changed()` is called from other thread.

hideki:
Issue for `Replication.status`
https://github.com/couchbase/couchbase-lite-java-core/issues/1167

PR #913, opened 2016-06-24
Fixed https://github.com/couchbase/couchbase-lite-java-core/issues/1289

hideki:
SecureTokenStore does not work with  Android M/N and there is encrypting data size limitation for API 18 - 22

- Implement TokenStore for Android M and N using AES.
- For API between 18 and 22, there is encrypting data size limitation (256 bytes). Applied different approach. Store RSA key for encrypting symmetric key that encrypts token. Encrypted symmetric key and encrypted token are stored in application storage.


pasin:
Can we use `com.couchbase.lite.support.security.SymmetricKey` class to do AES Key Generation for RSASecureTokenStore, encryption and decryption?

src/main/java/com/couchbase/lite/auth/Utils.java
```diff
@@ -0,0 +1,58 @@
+package com.couchbase.lite.auth;
```
pasin: Would it be better to have all Utils classes at the same place at `com/couchbase/lite/util/` package?

pasin: Need the license header as well.

hideki:
The class name of `SymmetricKey` is misleading. It does encryption and decryption. We could use SymmetricKey. But I'd like to change class name in the future.

pasin:
Actually the name is not misleading. I would suggest to keep the same name. Symmetric-Key is the name of the cryptography that uses the same key for both encrypting and decrypting the data. Also the class name is inline with the iOS.

hideki:
Personally, I prefer `SymmetricKeyAlgorithm`. The name ending with `Key` makes me think it is a  just key without encryption/decryption functionality. I will not change the name.

pasin:
Totally understand.

hideki:
`SecretKey` from Android `KeyStore` returns `null` for `getEncode()`. see http://stackoverflow.com/questions/32141884/encoded-text-of-secretkey-from-keystore-is-null-in-android-m
Without modifying SymmetricKey, it is not suitable for Android M and N solution.
I will use SymmetricKey only for Android API 18-22 solution, not for Android M/N solution.

pasin:
I will work on chaning the SymmetricKey class later.

PR #1019, opened 2016-10-18
Commented out two unit tests that MockServer returns 406. It should n…

hideki:
…ot be.

See https://github.com/couchbase/couchbase-lite-java-core/pull/1485


hideki:
These disabled test cases should be reviewed by ticket #1459.

PR #1092, opened 2017-03-08
Project structure, API interface for CRUD, DatabaseTest unit test

hideki:
- Couchbase Lite Android/Java 2.x project structure
- LiteCore as a submodule
- Database and Document API interfaces (from spec, without implementatoin) 
- DatabaseTest unit test


pasin:
I just noticed after merged that there is no .gitignore yet and .idea is also committed. Can you fix this?

PR #1105, opened 2017-03-21
Implementations for DB004

hideki:
Implemented Database, Document and Properties class and DatabaseTest and DocumentTest which are required for DB004

### DIFF
--- a/shared/src/main/java/com/couchbase/lite/PropertiesImpl.java
+++ b/shared/src/main/java/com/couchbase/lite/PropertiesImpl.java
@@ -2,12 +2,18 @@ package com.couchbase.lite;
 
 import com.couchbase.lite.internal.support.DateUtils;
 import com.couchbase.litecore.fleece.FLDict;
+import com.couchbase.litecore.fleece.FLValue;
 
 import java.util.Date;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+
+import static com.couchbase.litecore.fleece.FLConstants.FLValueType.kFLArray;
+import static com.couchbase.litecore.fleece.FLConstants.FLValueType.kFLBoolean;
+import static com.couchbase.litecore.fleece.FLConstants.FLValueType.kFLNumber;
+import static com.couchbase.litecore.fleece.FLConstants.FLValueType.kFLString;
 
 class PropertiesImpl implements Properties {
 
@@ -58,26 +64,65 @@ class PropertiesImpl implements Properties {
 
     @Override
     public String getString(String key) {
-        if (properties != null)
-            return (String) properties.get(key);
-        else
-            return root == null ? null : root.get(key).asString();
+        if (properties != null) {
+            Object obj = properties.get(key);
+            if (obj != null && obj instanceof String)
+                return (String) obj;
+            else
+                return null;
+        } else {
+            if (root == null)
+                return null;
+            else {
+                FLValue flvalue = root.get(key);
+                if (flvalue.getType() == kFLString)
+                    return flvalue.asString();
+                else
+                    return null;
+            }
+        }
     }
 
     @Override
     public int getInt(String key) {
-        if (properties != null)
-            return (int) properties.get(key);
-        else
-            return root == null ? 0 : root.get(key).asInt();
+        if (properties != null) {
+            Object obj = properties.get(key);
+            if (obj != null && obj instanceof Number)
+                return ((Number) obj).intValue();
+            else
+                return 0;
+        } else {
+            if (root == null)
+                return 0;
+            else {
+                FLValue flvalue = root.get(key);
+                if (flvalue.getType() == kFLNumber)
+                    return flvalue.asInt();
+                else
+                    return 0;
+            }
+        }
     }
 
     @Override
     public float getFloat(String key) {
-        if (properties != null)
-            return (float) properties.get(key);
-        else
-            return root == null ? 0.0F : root.get(key).asFloat();
+        if (properties != null) {
+            Object obj = properties.get(key);
+            if (obj != null && obj instanceof Number)
+                return ((Number) obj).floatValue();
+            else
+                return 0.0F;
+        } else{
+            if (root == null)
+                return 0.0F;
+            else {
+                FLValue flvalue = root.get(key);
+                if (flvalue.getType() == kFLNumber)
+                    return flvalue.asFloat();
+                else
+                    return 0.0F;
+            }
+        }
     }
 
     @Override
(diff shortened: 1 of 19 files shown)


shared/src/main/java/com/couchbase/lite/PropertiesImpl.java
```diff
@@ -82,67 +102,86 @@ public boolean getBoolean(String key) {
...
+        if (properties != null)
+            return DateUtils.fromJson((String) properties.get(key));
+        else
+            return root == null ? null : DateUtils.fromJson(root.get(key).asString());
     }
 
     @Override
     public List<Object> getArray(String key) {
         if (properties != null)
             return (List<Object>) properties.get(key);
```
pasin: If the value is not List, will it throw ClassCastException?

shared/src/main/java/com/couchbase/lite/PropertiesImpl.java
```diff
@@ -65,10 +73,22 @@ public int getInt(String key) {
     }
 
     @Override
-    public double getDouble(String key) {
+    public float getFloat(String key) {
         if (properties != null)
-            return (double) properties.get(key);
+            return (float) properties.get(key);
```
pasin: If the value is not number such as String or Map, will it throw the ClassCastException when casting to float?

shared/src/main/java/com/couchbase/lite/PropertiesImpl.java
```diff
@@ -82,67 +102,86 @@ public boolean getBoolean(String key) {
...
+        // TODO: DB005
         return null;
     }
 
     @Override
     public Date getDate(String key) {
-        // TODO: DB004
-        return null;
+        if (properties != null)
+            return DateUtils.fromJson((String) properties.get(key));
```
pasin: Same as above if properties.get(key) is not String, will this work?

shared/src/main/java/com/couchbase/lite/PropertiesImpl.java
hideki: As you wrote, should do type check.

shared/src/main/java/com/couchbase/lite/PropertiesImpl.java
hideki: same

shared/src/main/java/com/couchbase/lite/PropertiesImpl.java
hideki: same

shared/src/main/java/com/couchbase/lite/PropertiesImpl.java
```diff
@@ -58,26 +64,65 @@ public Object get(String key) {
...
+                return 0;
+            else {
+                FLValue flvalue = root.get(key);
+                if (flvalue.getType() == kFLNumber)
+                    return flvalue.asInt();
+                else
+                    return 0;
+            }
+        }
     }
```
pasin: Can you sync the implementation with the iOS to make the code more concise?
* iOS has numberProperty method.
* Utilize FLValue_AsXXX. I believe these are C methods.

hideki: Regarding the second comment, current Obj-C implementation does not check the FLValue type. It might return unexpected value

```Obj-C
- (double) doubleForKey: (NSString*)key {
    if (_properties)
        return [numberProperty(_properties, key) doubleValue];
    else
        return FLValue_AsDouble([self fleeceValueForKey: key]);
}

- (float) floatForKey: (NSString*)key {
    if (_properties)
        return [numberProperty(_properties, key) floatValue];
    else
        return FLValue_AsFloat([self fleeceValueForKey: key]);
}

- (NSInteger) integerForKey: (NSString*)key {
    if (_properties)
        return [numberProperty(_properties, key) integerValue];
    else
        return (NSInteger)FLValue_AsInt([self fleeceValueForKey: key]);
}
```

```C
bool FLValue_AsBool(FLValue v)                  {return v && v->asBool();}
int64_t FLValue_AsInt(FLValue v)                {return v ? v->asInt() : 0;}
uint64_t FLValue_AsUnsigned(FLValue v)          {return v ? v->asUnsigned() : 0;}
float FLValue_AsFloat(FLValue v)                {return v ? v->asFloat() : 0.0;}
double FLValue_AsDouble(FLValue v)              {return v ? v->asDouble() : 0.0;}
FLSlice FLValue_AsString(FLValue v)             {return v ? external(v->asString()) : kFLSliceNull;}
FLSlice FLValue_AsData(FLValue v)               {return v ? external(v->asData()) : kFLSliceNull;}
FLArray FLValue_AsArray(FLValue v)              {return v ? v->asArray() : nullptr;}
FLDict FLValue_AsDict(FLValue v)                {return v ? v->asDict() : nullptr;}
```

hideki: About the first comment:

If casting to NSNumber without checking type of FLValue, `numberProperty()` might return unexpected value?
```
static inline NSNumber* numberProperty(NSDictionary* dict, NSString* key) {
    return $castIf(NSNumber, dict[key]);
}
```

pasin: This method is for checking the value from dictionary not FLValue.

hideki: I could write the method to concise the codes.

pasin: IIRC, v->asInt() will check the type.

pasin: For -dateForKey:, iOS might have a bug. It should get value as String first before parsing to the date object.

hideki: Now I got your point.

In case of `asBool()` method, it returns true for any string value. 
But Java, Boolean.valueof(randam String) return false.
It is reason that I added type check before returning value.

In case of `asInt()` it returns always 0 for non number values.

```
    bool Value::asBool() const noexcept {
        switch (tag()) {
            case kSpecialTag:
                return tinyValue() == kSpecialValueTrue;
            case kShortIntTag:
            case kIntTag:
            case kFloatTag:
                return asInt() != 0;
            default:
                return true;
        }
    }
```

hideki: Except `asBool()`, any asXXX() methods returns the expected values.

shared/src/main/java/com/couchbase/lite/PropertiesImpl.java
```diff
@@ -177,43 +124,22 @@ public Blob getBlob(String key) {
...
-                    return DateUtils.fromJson(flvalue.asString());
-                else
-                    return null;
-            }
+        if (properties != null)
+            return DateUtils.fromJson(cast(properties.get(key), String.class));
+        else {
+            FLValue flvalue = getValueFromRoot(key);
+            return flvalue == null || flvalue.getType() != kFLString
+                    ? null : DateUtils.fromJson(flvalue.asString());
```
pasin: This can be done by `return DateUtils.fromJson(getString(key))` so that the get string code doesn't get repeated here.

shared/src/main/java/com/couchbase/lite/PropertiesImpl.java
pasin: I just saw the asBool() comment. There are several implementation based on language or platform when determining boolean value. I think we should stick with the lite core interpretation for cross-platform consistency. I remembered fixing getBoolean() method to inline with the lite core for iOS in DB2.

hideki: CBL Java DB004 is based on iOS DB1. I will revisit this during next dev cycle.

PR #1171, opened 2017-05-20
Fixed #1157 - 2.0: Document.remove(key) is not functioning.

### LINKED ISSUE #1157: 2.0: Document.remove(key) is not functioning.
- `Dictionary.remove(String key)`

hideki:
- Implemented Dictionary.remove(String key) by RemovedValue substitution.

### DIFF
--- a/shared/src/main/java/com/couchbase/lite/Dictionary.java
+++ b/shared/src/main/java/com/couchbase/lite/Dictionary.java
@@ -1,13 +1,16 @@
 package com.couchbase.lite;
 
+import com.couchbase.lite.internal.document.RemovedValue;
 import com.couchbase.lite.internal.support.DateUtils;
 import com.couchbase.litecore.fleece.FLEncoder;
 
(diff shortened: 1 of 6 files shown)


shared/src/main/java/com/couchbase/lite/Dictionary.java
```diff
@@ -80,7 +83,13 @@ public Dictionary set(Map<String, Object> dictionary) {
         }
 
         // Marked the key as removed by setting the value to kRemovedValue:
-        // TODO
+        Map<String, Object> backingData = super.toMap();
+        if (backingData != null) {
+            for (Map.Entry<String, Object> entry : backingData.entrySet()) {
```
pasin: Doesn't Map iterator used entry set by default?

hideki: java `Map` interface does not extend `Iterable`.

Issue #1170, opened 2017-05-19
Array and Dictionary should throw ConcurrentModificationException if Array or Dictionary is modified.


hideki:
Map/HashMap/List/ArrayList behavior.

`ConcurrentModificationException` is thrown by `Iterator.next()`
#### Sample code
```java
package self.hideki;

import java.util.ArrayList;
import java.util.Iterator;
import java.util.List;

public class Main {
    public static void main(String[] args) {
        int count = 0;
	    List<Object> list = new ArrayList<Object>();
	    list.add(1);
	    list.add(2);
        Iterator<Object> itr = list.iterator();
        while(itr.hasNext()){
            System.out.println(itr.next());
            if(count++ == 0)
                list.add(3);
        }
    }
}
```

#### Output
```
1
Exception in thread "main" java.util.ConcurrentModificationException
	at java.util.ArrayList$Itr.checkForComodification(ArrayList.java:901)
	at java.util.ArrayList$Itr.next(ArrayList.java:851)
	at self.hideki.Main.main(Main.java:16)
```

Issue #1161, opened 2017-05-16
2.0: Double value which is stored in fleece is returned as Float.


hideki:
### Test 
- ArrayTest.testSetGetFloatNumbers()

hideki:
```c
    void Encoder::writeDouble(double n) {
        throwIf(std::isnan(n), InvalidData, "Can't write NaN");
        if (n == (int64_t)n) {
            return writeInt((int64_t)n);
        } else if (n == (float)n) {
            return _writeFloat((float)n);
        } else {
            littleEndianDouble swapped = n;
            uint8_t buf[2 + sizeof(swapped)];
            buf[0] = 0x08; // 'double' size flag
            buf[1] = 0;
            memcpy(&buf[2], &swapped, sizeof(swapped));
            writeValue(kFloatTag, buf, sizeof(buf));
        }
    }
```

hideki:
Fix is in https://github.com/couchbase/couchbase-lite-android/tree/feature/dev_db008 branch.

hideki:
PR: https://github.com/couchbase/couchbase-lite-android/pull/1180

hideki:
PR: https://github.com/couchbase/couchbase-lite-android/pull/1180

Issue #1160, opened 2017-05-16
2.0: Double/Float without decimal are returned as Integer


hideki:
### Test
ArrayTest.testSetGetFloatNumbers

hideki:
Fleece stores floating value without floating part such as 1.0 or 10.00 as `int` value. getObject() and getNumber() will return `Long`.

```c
    /** Writes a 64-bit floating point number to an encoder.
        (Note: as an implementation detail, if the number has no fractional part and can be
        represented exactly as an integer, it'll be encoded as an integer to save space. This is
        transparent to the reader, since if it requests the value as a float it'll be returned
        as floating-point.) */
    bool FLEncoder_WriteDouble(FLEncoder, double);
```

```c
    void Encoder::writeDouble(double n) {
        throwIf(std::isnan(n), InvalidData, "Can't write NaN");
        if (n == (int64_t)n) {
            return writeInt((int64_t)n);
        } else if (n == (float)n) {
            return _writeFloat((float)n);
        } else {
            littleEndianDouble swapped = n;
            uint8_t buf[2 + sizeof(swapped)];
            buf[0] = 0x08; // 'double' size flag
            buf[1] = 0;
            memcpy(&buf[2], &swapped, sizeof(swapped));
            writeValue(kFloatTag, buf, sizeof(buf));
        }
    }
```

hideki:
Fix is in https://github.com/couchbase/couchbase-lite-android/tree/feature/dev_db008 branch.

hideki:
PR: https://github.com/couchbase/couchbase-lite-android/pull/1180

Issue #1158, opened 2017-05-16
2.0: getLong/setLong does not work as expected


hideki:
- DocumentTest.testSetGetMinMaxNumbers
- ArrayTest.testSetGetMinMaxNumbers
- ArrayTest.testSetGetFloatNumbers

hideki:
Fix is in https://github.com/couchbase/couchbase-lite-android/tree/feature/dev_db008 branch.

hideki:
PR: https://github.com/couchbase/couchbase-lite-android/pull/1180

PR #1180, opened 2017-05-24
2.0 API updates and bug fixes

### LINKED ISSUE #1174: 2.0: API update
- [x] Add Database.count and Database.contain(id)
- [x] Add ReadOnlyDictionary.keys
- [x]  Add Query.explain() and Query.profile()
- [x] Add some comment specific to Android platform on DatabaseConfiguration and Database constructor.

Please see https://github.com/couchbaselabs/couchbase-lite-apiv2/commit/9a4222109366e192863dd02005cb58e158352183

hideki:
#1174 - 2.0: API update
  - Implemented DatabaseConfiguration(Context) for Android
  - Implemented Dictionary.getKeys() instead of getAllKeys()
  - Removed Database(String) consturctor which is inrelebant for Android.
  - Implemented Query.explain()
  - Ported two Unit Test cses for Document Change Listener
  - Ported testRemoveDictionary() unit test. And also fixed the bug which was found by the test

#1158 - 2.0: getLong/setLong does not work as expected
  - Casting to int from long cause data loss. Fixed this.

#1160 - 2.0: Double/Float without decimal are returned as Integer
  - Fleece stores double or float value without decimal fraction as the integer value. Fixed the test cases

#1161 - 2.0: Double value which is stored in fleece is returned as Float.
  - Fleece stores the double value, which is same as the float value, as the float value. Fixed the test cases

#1170 - Array and Dictionary should throw ConcurrentModificationException if Array or Dictionary is modified.
 - Check Array/Dictionary updates during iteration.

Note: For details, please refer each commit?

### DIFF
--- a/android/CouchbaseLite/src/androidTest/java/com/couchbase/lite/ArrayTest.java
+++ b/android/CouchbaseLite/src/androidTest/java/com/couchbase/lite/ArrayTest.java
@@ -8,7 +8,7 @@ import org.junit.rules.ExpectedException;
 
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.HashMap;
+import java.util.TreeMap;
 import java.util.List;
 import java.util.Map;
 
@@ -156,16 +156,16 @@ public class ArrayTest extends BaseTest {
                 assertEquals(true, a.getObject(0));
                 assertEquals(false, a.getObject(1));
                 assertEquals("string", a.getObject(2));
-                assertEquals(0, a.getObject(3));
-                assertEquals(1, a.getObject(4));
-                assertEquals(-1, a.getObject(5));
+                assertEquals(0, ((Number)a.getObject(3)).intValue());
+                assertEquals(1, ((Number)a.getObject(4)).intValue());
+                assertEquals(-1, ((Number)a.getObject(5)).intValue());
                 assertEquals(1.1, a.getObject(6));
                 assertEquals(kArrayTestDate, a.getObject(7));
                 assertEquals(null, a.getObject(8));
 
                 // dictionary
                 Dictionary subdict = (Dictionary) a.getObject(9);
-                Map<String, Object> expectedMap = new HashMap<>();
+                Map<String, Object> expectedMap = new TreeMap<>();
                 expectedMap.put("name", "Scott Tiger");
                 assertEquals(expectedMap, subdict.toMap());
 
@@ -212,16 +212,16 @@ public class ArrayTest extends BaseTest {
                 assertEquals(true, a.getObject(12 + 0));
                 assertEquals(false, a.getObject(12 + 1));
                 assertEquals("string", a.getObject(12 + 2));
-                assertEquals(0, a.getObject(12 + 3));
-                assertEquals(1, a.getObject(12 + 4));
-                assertEquals(-1, a.getObject(12 + 5));
+                assertEquals(0, ((Number)a.getObject(12 + 3)).intValue());
+                assertEquals(1, ((Number)a.getObject(12 + 4)).intValue());
+                assertEquals(-1, ((Number)a.getObject(12 + 5)).intValue());
                 assertEquals(1.1, a.getObject(12 + 6));
                 assertEquals(kArrayTestDate, a.getObject(12 + 7));
                 assertEquals(null, a.getObject(12 + 8));
 
                 // dictionary
                 Dictionary subdict = (Dictionary) a.getObject(12 + 9);
-                Map<String, Object> expectedMap = new HashMap<>();
+                Map<String, Object> expectedMap = new TreeMap<>();
                 expectedMap.put("name", "Scott Tiger");
                 assertEquals(expectedMap, subdict.toMap());
 
@@ -263,16 +263,16 @@ public class ArrayTest extends BaseTest {
                 assertEquals(true, a.getObject(0));
                 assertEquals(false, a.getObject(1));
                 assertEquals("string", a.getObject(2));
-                assertEquals(0, a.getObject(3));
-                assertEquals(1, a.getObject(4));
-                assertEquals(-1, a.getObject(5));
+                assertEquals(0, ((Number)a.getObject(3)).intValue());
+                assertEquals(1, ((Number)a.getObject(4)).intValue());
+                assertEquals(-1, ((Number)a.getObject(5)).intValue());
                 assertEquals(1.1, a.getObject(6));
                 assertEquals(kArrayTestDate, a.getObject(7));
                 assertEquals(null, a.getObject(8));
 
                 // dictionary
                 Dictionary subdict = (Dictionary) a.getObject(9);
-                Map<String, Object> expectedMap = new HashMap<>();
+                Map<String, Object> expectedMap = new TreeMap<>();
                 expectedMap.put("name", "Scott Tiger");
                 assertEquals(expectedMap, subdict.toMap());
 
@@ -523,9 +523,9 @@ public class ArrayTest extends BaseTest {
                 assertNull(a.getNumber(0));
                 assertNull(a.getNumber(1));
                 assertNull(a.getNumber(2));
-                assertEquals(0, a.getNumber(3));
-                assertEquals(1, a.getNumber(4));
-                assertEquals(-1, a.getNumber(5));
+                assertEquals(0, ((Number)a.getNumber(3)).intValue());
+                assertEquals(1, ((Number)a.getNumber(4)).intValue());
+                assertEquals(-1, ((Number)a.getNumber(5)).intValue());
                 assertEquals(1.1, a.getNumber(6));
                 assertNull(a.getNumber(7));
                 assertNull(a.getNumber(8));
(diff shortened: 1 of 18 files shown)


shared/src/main/java/com/couchbase/lite/ReadOnlyDictionary.java
```diff
@@ -99,7 +104,7 @@ public Number getNumber(String key) {
     @Override
     public int getInt(String key) {
         FLValue value = fleeceValue(key);
-        return value != null ? value.asInt() : 0;
+        return value != null ? (int)value.asInt() : 0;
```
pasin: Same here. It seems like asInt() doesn't return int?

android/CouchbaseLite/src/main/java/com/couchbase/lite/DatabaseConfiguration.java
```diff
@@ -0,0 +1,29 @@
+package com.couchbase.lite;
+
+import android.content.Context;
+
+public class DatabaseConfiguration extends BaseDatabaseConfiguration{
```
pasin: Formatting :)

shared/src/main/java/com/couchbase/lite/Dictionary.java
```diff
@@ -356,7 +366,7 @@ public Date getDate(String key) {
      */
     @Override
     public Map<String, Object> toMap() {
-        Map<String, Object> result = map != null ? new HashMap<>(map) : new HashMap<String, Object>();
+        Map<String, Object> result = map != null ? new TreeMap<>(map) : new TreeMap<String, Object>();
```
pasin: FYI, JSONObject is also used HashMap as a backing data.

https://android.googlesource.com/platform/dalvik/+/f2f7880a40aed1c8d9c27db49226e47396556aac/libcore/json/src/main/java/org/json/JSONObject.java

android/CouchbaseLite/src/androidTest/java/com/couchbase/lite/ArrayTest.java
```diff
@@ -701,9 +700,10 @@ public void testSetGetFloatNumbers() {
         save(doc, "array", array, new Validator<Array>() {
             @Override
             public void validate(Array a) {
-                // TODO: https://github.com/couchbase/couchbase-lite-android/issues/1160
-//                assertEquals(1.00, a.getObject(0));
-//                assertEquals(1.00, a.getNumber(0));
+                // NOTE: Number which has no floating part is stored as Integer.
+                //       This causes type difference between before and after storing data into the database.
+                assertEquals(1.00, ((Number)a.getObject(0)).doubleValue(), 0.0);
```
hideki: LiteCore (Fleece) stores 1.0 as 1 to save space.
```c
 void Encoder::writeDouble(double n) {
        throwIf(std::isnan(n), InvalidData, "Can't write NaN");
        if (n == (int64_t)n) {
            return writeInt((int64_t)n);
        } else if (n == (float)n) {
            return _writeFloat((float)n);
        } else {
            littleEndianDouble swapped = n;
            uint8_t buf[2 + sizeof(swapped)];
            buf[0] = 0x08; // 'double' size flag
            buf[1] = 0;
            memcpy(&buf[2], &swapped, sizeof(swapped));
            writeValue(kFloatTag, buf, sizeof(buf));
        }
    }
```

android/CouchbaseLite/src/androidTest/java/com/couchbase/lite/ArrayTest.java
```diff
@@ -156,16 +156,16 @@ public void validate(Array a) {
...
+                assertEquals(1, ((Number)a.getObject(4)).intValue());
+                assertEquals(-1, ((Number)a.getObject(5)).intValue());
                 assertEquals(1.1, a.getObject(6));
                 assertEquals(kArrayTestDate, a.getObject(7));
                 assertEquals(null, a.getObject(8));
 
                 // dictionary
                 Dictionary subdict = (Dictionary) a.getObject(9);
-                Map<String, Object> expectedMap = new HashMap<>();
+                Map<String, Object> expectedMap = new TreeMap<>();
```
hideki: TreeMap is easier for testing as the order of keys is guaranteed.
I can switch it back.

android/CouchbaseLite/src/androidTest/java/com/couchbase/lite/ArrayTest.java
```diff
@@ -716,9 +716,8 @@ public void validate(Array a) {
                 assertEquals(1.49F, a.getFloat(1), 0.0F);
                 assertEquals(1.49, a.getDouble(1), 0.0);
 
-                // TODO:
-//                assertEquals(1.50, a.getObject(2));
-//                assertEquals(1.50, a.getNumber(2));
+                assertEquals(1.50, ((Number)a.getObject(2)).doubleValue(), 0.0);
+                assertEquals(1.50, a.getNumber(2).doubleValue(), 0.0);
```
hideki: same reason with above. Some of the Double value is stored as Float.
1.50 on java is Double, and a.getObject() returns Float as Fleece store it as Float.

android/CouchbaseLite/src/androidTest/java/com/couchbase/lite/DictionaryTest.java
```diff
@@ -3,34 +3,35 @@
...
+import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
 
 public class DictionaryTest extends BaseTest {
     @Test
     public void testCreateDictionary() {
         Dictionary address = new Dictionary();
         assertEquals(0, address.count());
-        assertEquals(new HashMap<String, Object>(), address.toMap());
+        assertEquals(new TreeMap<String, Object>(), address.toMap());
```
hideki: I will switch it back to HashMap. And will consider another solution for map comparison.

android/CouchbaseLite/src/androidTest/java/com/couchbase/lite/ArrayTest.java
pasin: What is the rule to store as float or double?

android/CouchbaseLite/src/androidTest/java/com/couchbase/lite/DocumentTest.java
```diff
@@ -177,16 +177,16 @@ public void testCreateDocWithDict() {
         assertTrue(doc1b.exists());
         assertFalse(doc1b.isDeleted());
         assertEquals(doc1a.getId(), doc1b.getId());
-        assertEquals(dict, doc1b.toMap());
+        assertEquals(dict.toString(), doc1b.toMap().toString());
```
hideki: I will take a look Hamcrest's assertion

android/CouchbaseLite/src/main/java/com/couchbase/lite/DatabaseConfiguration.java
hideki: ok

android/CouchbaseLite/src/main/java/com/couchbase/lite/DatabaseConfiguration.java
```diff
@@ -0,0 +1,29 @@
...
+    private Context context;
+
+    public DatabaseConfiguration() {
```
hideki: `Database(String name)` is not allowed for Android. But `DatabaseConfiguration()` should be allowed because the user can set directory.

shared/src/main/java/com/couchbase/lite/Dictionary.java
```diff
@@ -97,6 +99,14 @@ public Dictionary set(Map<String, Object> dictionary) {
         return this;
     }
 
+    @Override
+    public List<String> getKeys() {
+        if (!changed)
+            return super.getKeys();
+        else
+            return allKeys();
```
hideki: Both Dictionary.allKeys() and ReadOnlyDictonary.getKeys() create new instance of `List<String>`. It is not necessary.

shared/src/main/java/com/couchbase/lite/ReadOnlyDictionary.java
hideki: Fleece has one type for integral numbers. It is Integer. 
Fleece's Integer is equivalent with Java Long. 2^64.

shared/src/main/java/com/couchbase/lite/ReadOnlyArray.java
```diff
@@ -88,7 +88,7 @@ public Number getNumber(int index) {
      */
     @Override
     public int getInt(int index) {
-        return fleeceValue(index).asInt();
+        return (int)fleeceValue(index).asInt();
```
hideki: Fleece Integer is equivalent with Java Long.
I implemented JNI to return long. Otherwise, Java could not support Long.MAX and Long.MIN.

shared/src/main/java/com/couchbase/lite/Dictionary.java
pasin: Could you double check allKeys() method of the Dictionary? I remembered that it does cache the keys.

shared/src/main/java/com/couchbase/lite/Dictionary.java
hideki: As TreeMap maintains the order of keys, I changed it to TreeMap instead of HashMap. I will switch back to HashMap.

android/CouchbaseLite/src/main/java/com/couchbase/lite/DatabaseConfiguration.java
pasin: I think we just don't allow. Otherwise we will need to check whether users set directory or not and throw runtime exception. In the future, if this becomes a concern of difficult to use, we can add back. What do you think?

hideki:
TODOs:
- Switch back TreeMap to HashMap
- Fix Map value comparison for unit test

hideki:
@pasin Can you review this PR again? Thanks!

pasin:
I have added my comment about copying keys and DatabaseConfiguration constructor. Can you review?

android/CouchbaseLite/src/main/java/com/couchbase/lite/DatabaseConfiguration.java
pasin: Can you review my comment above^?

android/CouchbaseLite/src/androidTest/java/com/couchbase/lite/ArrayTest.java
hideki: Depend on following logic
```c
void Encoder::writeDouble(double n) {
        throwIf(std::isnan(n), InvalidData, "Can't write NaN");
        if (n == (int64_t)n) {
            return writeInt((int64_t)n);
        } else if (n == (float)n) {
            return _writeFloat((float)n);
        } else {
            littleEndianDouble swapped = n;
            uint8_t buf[2 + sizeof(swapped)];
            buf[0] = 0x08; // 'double' size flag
            buf[1] = 0;
            memcpy(&buf[2], &swapped, sizeof(swapped));
            writeValue(kFloatTag, buf, sizeof(buf));
        }
    }
```

shared/src/main/java/com/couchbase/lite/Dictionary.java
hideki: I will check

pasin: Can you review my comment above ^?

hideki:
I did not read your latest comment. I will check Dictionary keys.

android/CouchbaseLite/src/main/java/com/couchbase/lite/DatabaseConfiguration.java
hideki: So you are saying `DatabaseConfiguration` constructor only support `DatabaseConfiguration(Context)` for Android? To simplify the API, this could be a reasonable solution.

pasin: Yes.

hideki:
As you said, keys variable is cached.

hideki:
@pasin Could you please re-review PR? Thanks!

hideki:
Made a new ticket for `Hamcrest matchers` for unit tests
https://github.com/couchbase/couchbase-lite-android/issues/1181

shared/src/main/java/com/couchbase/lite/Dictionary.java
```diff
@@ -103,9 +103,9 @@ public Dictionary set(Map<String, Object> dictionary) {
     @Override
     public List<String> getKeys() {
         if (!changed)
-            return super.getKeys();
+            return new ArrayList(super.getKeys());
```
pasin: There is no needs to create a new ArrayList here as super.getKeys() already returns a copy.

hideki:
@pasin Thank you for catching the problem.

PR #1194, opened 2017-05-27
Fixed #1187 - 2.0: Review failing DatabaseTest unit tests

### LINKED ISSUE #1187: 2.0: Review failing DatabaseTest unit tests

hideki:
NOTE: Please review Database.java

- As temporary solution, If c4Database is already closed, CBL throws CouchbaseLiteException with DBClosed (450) code. NOTE: This is currently temporary.
- Database.internal() check if c4db is null. If it is null, throw CouchbaseLiteException.
- Reviewed other unit test failures with DatabaseTest.

### DIFF
--- a/android/CouchbaseLite/src/androidTest/java/com/couchbase/lite/ConflictTest.java
+++ b/android/CouchbaseLite/src/androidTest/java/com/couchbase/lite/ConflictTest.java
@@ -94,11 +94,11 @@ public class ConflictTest extends BaseTest {
             @Override
             public void run() {
                 try {
-                    com.couchbase.litecore.Document trickey = db.internal().getDocument(docID, true);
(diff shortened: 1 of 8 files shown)


pasin:
I'm not sure throwing CouchbaseLiteException is the cross platform solution as the other platforms doesn't implement CouchbaseLiteException as runtime exception. For example, Objective-C needs to pass out NSError pointer and Swift throws non-runtime exception.

hideki:
@pasin Please review PR. Again.

#### Changes: 

In case the database is closed, the database operations throw `IllegalStateException` instead of `CouchbaseLiteException`.

Definition of `IllegalStateExcepition`

> Signals that a method has been invoked at an illegal or inappropriate time. In other words, the Java environment or Java application is not in an appropriate state for the requested operation.

Also replaced name if `internal()` method to `getC4Database()`

PR #1221, opened 2017-06-16
Fixed #1205 - 2.0: LiveQuery

### LINKED ISSUE #1205: 2.0: LiveQuery

hideki:
- Implemented LiveQuery
- Added Unit tests

### DIFF
--- a/shared/src/main/java/com/couchbase/lite/LiveQuery.java
+++ b/shared/src/main/java/com/couchbase/lite/LiveQuery.java
@@ -1,0 +1,172 @@
+package com.couchbase.lite;
+
+import android.os.Handler;
+import android.os.Looper;
+
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * A Query subclass that automatically refreshes the result rows every time the database changes.
+ */
+public class LiveQuery implements DatabaseChangeListener {
+
+    //---------------------------------------------
+    // static variables
+    //---------------------------------------------
+    private final static String TAG = Log.QUERY;
+    private final static long kDefaultLiveQueryUpdateInterval = 200; // 0.2sec (200ms)
+
+    //---------------------------------------------
+    // member variables
+    //---------------------------------------------
+
+    private Set<LiveQueryChangeListener> queryChangeListener;
+    private Query query;
+    private ResultSet resultSet;
+    private boolean observing;
+    private boolean willUpdate;
+    private boolean forceReload;
+    private long lastUpdatedAt;
+
+    //---------------------------------------------
+    // Constructors
+    //---------------------------------------------
+
+    /* package */ LiveQuery(Query query) {
+        this.query = query;
+        queryChangeListener = Collections.synchronizedSet(new HashSet<LiveQueryChangeListener>());
+        resultSet = null;
+        observing = false;
+        willUpdate = false;
+        forceReload = false;
+        lastUpdatedAt = 0L;
+    }
+
+    //---------------------------------------------
+    // API - public methods
+    //---------------------------------------------
+
+    /**
+     * Starts observing database changes. The .rows property will now update automatically.
+     * (You usually don't need to call this yourself, since accessing or observing the getRows()
+     * property will call -start for you.)
+     */
+    public void run() {
+        if (!observing) {
+            observing = true;
+            query.getDatabase().addChangeListener(this);
+            update();
+        }
+    }
+
+    /**
+     * Stops observing database changes. Calling start() or getRows() will restart it.
+     */
+    public void stop() {
+        if (observing) {
+            observing = false;
+            query.getDatabase().removeChangeListener(this);
+        }
+        willUpdate = false; // cancels the delayed update started by -databaseChanged
+    }
+
+    public void addLiveQueryChangeListener(LiveQueryChangeListener listener) {
+        queryChangeListener.add(listener);
+        if (queryChangeListener.size() == 1)
+            run();
+    }
+
+    public void removeLiveQueryChangeListener(LiveQueryChangeListener listener) {
+        queryChangeListener.remove(listener);
+        if (queryChangeListener.size() == 0)
+            stop();
+    }
+
+    //---------------------------------------------
+    // Implementation of DatabaseChangeListener
+    //---------------------------------------------
+    @Override
+    public void changed(DatabaseChange change) {
+        if (willUpdate)
+            return; // Already a pending update scheduled
+
+        // Schedule an update, respecting the updateInterval:
+        long updateDelay = lastUpdatedAt + kDefaultLiveQueryUpdateInterval - System.currentTimeMillis();
+        updateDelay = Math.max(0, Math.min(this.kDefaultLiveQueryUpdateInterval, updateDelay));
+        update(updateDelay);
+    }
+
+    //---------------------------------------------
+    // Package level access
+    //---------------------------------------------
+
+    /**
+     * The current query results; this updates as the database changes.
+     * Its value will be null until the initial asynchronous query finishes.
+     * <p>
+     * NOTE: for unit test
+     */
+    /*package*/ ResultSet getResultSet() {
+        run();
+        return resultSet;
+    }
+
+    //---------------------------------------------
+    // Private (in class only)
+    //---------------------------------------------
+
+    /**
+     * @param delay millisecond
+     */
+    private void update(long delay) {
+        if (willUpdate)
+            return; // Already a pending update scheduled
+
+        willUpdate = true;
+        // create one doc
+        new Handler(Looper.getMainLooper())
+                .postDelayed(new Runnable() {
+                    @Override
+                    public void run() {
+                        update();
+                    }
+                }, delay); // ms
+    }
+
+    private void update() {
+        // TODO: Make this asynchronous (as in 1.x)
+        try {
+            Log.i(TAG, "%s: Querying...", this);
+            ResultSet oldResultSet = resultSet;
+            ResultSet newResultSet;
+            if (oldResultSet == null || !oldResultSet.isValidEnumerator() || forceReload)
+                newResultSet = query.run();
+            else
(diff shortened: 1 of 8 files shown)


shared/src/main/java/com/couchbase/lite/LiveQuery.java
```diff
@@ -0,0 +1,172 @@
...
+    private boolean willUpdate;
+    private boolean forceReload;
+    private long lastUpdatedAt;
+
+    //---------------------------------------------
+    // Constructors
+    //---------------------------------------------
+
+    /* package */ LiveQuery(Query query) {
+        this.query = query;
```
pasin: I think we should copy the query.

shared/src/main/java/com/couchbase/lite/LiveQuery.java
```diff
@@ -0,0 +1,172 @@
...
+     */
+    public void stop() {
+        if (observing) {
+            observing = false;
+            query.getDatabase().removeChangeListener(this);
+        }
+        willUpdate = false; // cancels the delayed update started by -databaseChanged
+    }
+
+    public void addLiveQueryChangeListener(LiveQueryChangeListener listener) {
```
pasin: Should be addChangeListener.

shared/src/main/java/com/couchbase/lite/LiveQuery.java
```diff
@@ -0,0 +1,172 @@
...
+        queryChangeListener.add(listener);
+        if (queryChangeListener.size() == 1)
+            run();
+    }
+
+    public void removeLiveQueryChangeListener(LiveQueryChangeListener listener) {
```
pasin: Should be removeChangeListener.

shared/src/main/java/com/couchbase/lite/LiveQuery.java
hideki: A copy of Query instance will be created in the `ToLive()` method.

hideki:
Applied the feedbacks. Please review this again. Thanks!

shared/src/main/java/com/couchbase/lite/LiveQuery.java
```diff
@@ -0,0 +1,172 @@
...
+    // Package level access
+    //---------------------------------------------
+
+    /**
+     * The current query results; this updates as the database changes.
+     * Its value will be null until the initial asynchronous query finishes.
+     * <p>
+     * NOTE: for unit test
+     */
+    /*package*/ ResultSet getResultSet() {
```
pasin: I don't see this one get used on LiveQuery class. Can this be removed?

shared/src/main/java/com/couchbase/lite/LiveQuery.java
```diff
@@ -0,0 +1,172 @@
...
+    //---------------------------------------------
+    // API - public methods
+    //---------------------------------------------
+
+    /**
+     * Starts observing database changes. The .rows property will now update automatically.
+     * (You usually don't need to call this yourself, since accessing or observing the getRows()
+     * property will call -start for you.)
+     */
+    public void run() {
```
pasin: Calling run again should force to rerun the query.

shared/src/main/java/com/couchbase/lite/LiveQuery.java
```diff
@@ -0,0 +1,172 @@
...
+            observing = false;
+            query.getDatabase().removeChangeListener(this);
+        }
+        willUpdate = false; // cancels the delayed update started by -databaseChanged
+    }
+
+    public void addChangeListener(LiveQueryChangeListener listener) {
+        queryChangeListener.add(listener);
+        if (queryChangeListener.size() == 1)
+            run();
```
pasin: This is not part of the spec. What do you think about this in general?

shared/src/main/java/com/couchbase/lite/LiveQuery.java
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    public void removeChangeListener(LiveQueryChangeListener listener) {
+        queryChangeListener.remove(listener);
+        if (queryChangeListener.size() == 0)
+            stop();
```
pasin: This is also not part of the spec. Should we include this in spec?

shared/src/main/java/com/couchbase/lite/LiveQuery.java
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    private void sendNotification(LiveQueryChange change) {
+        synchronized (queryChangeListener) {
+            for (LiveQueryChangeListener listener : queryChangeListener) {
+                listener.changed(change);
+            }
+        }
+    }
+}
```
pasin: Should also remove the listener to the database when the object is collected as well.

shared/src/main/java/com/couchbase/lite/LiveQuery.java
hideki: This implementation does not work well with the behavior required of `run()` method.

If we add this, we should change `run()` method as a non-public method.

shared/src/main/java/com/couchbase/lite/LiveQuery.java
hideki: same with above

shared/src/main/java/com/couchbase/lite/LiveQuery.java
hideki: This is used from unit test only. I will remove this.

hideki:
Please review this PR again. Thanks!

shared/src/main/java/com/couchbase/lite/LiveQuery.java
```diff
@@ -0,0 +1,181 @@
...
+        this.lastUpdatedAt = 0L;
+    }
+
+    //---------------------------------------------
+    // API - public methods
+    //---------------------------------------------
+
+    /**
+     * Starts observing database changes. The .rows property will now update automatically.
+     * NOTE: Calling run method will (re)start the live query.
```
pasin: Please update the comment.

shared/src/main/java/com/couchbase/lite/LiveQuery.java
```diff
@@ -0,0 +1,181 @@
...
+     */
+    public void run() {
+        observing = true;
+        releaseResultSet();
+        query.getDatabase().addChangeListener(this);
+        update();
+    }
+
+    /**
+     * Stops observing database changes. Calling start() or getRows() will restart it.
```
pasin: Need update.

shared/src/main/java/com/couchbase/lite/LiveQuery.java
```diff
@@ -0,0 +1,181 @@
...
+        // TODO: Make this asynchronous (as in 1.x)
+
+        if (!observing)
+            return;
+
+        try {
+            Log.i(TAG, "%s: Querying...", this);
+            ResultSet oldResultSet = resultSet;
+            ResultSet newResultSet;
+            if (oldResultSet == null || !oldResultSet.isValidEnumerator())
```
pasin: What is the case that the oldResultSet will be invalid? I'm wondering if the enumerator is still valid, if the app iterates to all results?

shared/src/main/java/com/couchbase/lite/LiveQuery.java
```diff
@@ -0,0 +1,181 @@
...
+
+    @Override
+    protected void finalize() throws Throwable {
+        stop();
+        super.finalize();
+    }
+
+    //---------------------------------------------
+    // Package level access
+    //---------------------------------------------
```
pasin: Do you prefer to keep this comment when there are no any package level items?

shared/src/main/java/com/couchbase/lite/LiveQuery.java
```diff
@@ -0,0 +1,181 @@
...
+            if (newResultSet != null) {
+                if (oldResultSet != null)
+                    Log.i(TAG, "%s: Changed!", this);
+                resultSet = newResultSet;
+                sendNotification(new LiveQueryChange(this, resultSet, null));
+            } else {
+                Log.i(TAG, "%s: ...no change", this);
+            }
+        } catch (CouchbaseLiteException e) {
+            sendNotification(new LiveQueryChange(null, e));
```
pasin: The first parameter shouldn't be null.

shared/src/main/java/com/couchbase/lite/LiveQueryChange.java
```diff
@@ -0,0 +1,44 @@
...
+
+    /*package*/ LiveQueryChange(LiveQuery query, ResultSet rows) {
+        this(query, rows, null);
+    }
+
+    /*package*/ LiveQueryChange(LiveQuery query, Throwable error) {
+        this(query, null, error);
+    }
+
+    /*package*/ LiveQueryChange(LiveQuery query, ResultSet rows, Throwable error) {
```
pasin: I think having on this one is enough.

shared/src/main/java/com/couchbase/lite/LiveQuery.java
hideki: C4QueryEnumerator.next() of native code closes enumerator automatically if it reaches the end of the enumerator. So it is why it checks if the enumerator is still valid.

hideki:
Please re-review this PR again. Thanks!

shared/src/main/java/com/couchbase/lite/LiveQuery.java
pasin: If it closes automatically, it is not correct to check here as it has a side effect.
For example, 
1. User iterates through the ResultSet -> enumerator closes automatically.
2. There is a database change event but the event doesn't cause the query result to be changed.
3. The update() method is called due to step 2. As the enumerator is not valid anymore, the query is re-run and returns the same old result.

hideki:
Thank you for catching the problem. It is reproducible with the unit test. Please re-review

PR #1259, opened 2017-07-08
Fixed #1257 - 2.0 Query -  Projection, Functions, GroupBy, Having, and API updates

### LINKED ISSUE #1257: 2.0 Query -  Projection (SelectResult), Aggregate function, GroupBy, Having and Limit
- [x] Projection (ResultSet)
- [x] Aggregate functions
  - [x] avg
  - [x] count
  - [x] min
  - [x] max
  - [x] sum
- [x] GroupBy
- [x] Having
- [ ] ~~Limit~~

API updates for followings
- [x] Join
- [x] OrderBy
- [x] GroupBy

hideki:
Query:
- Projection (SelectResult)
- Aggregate Functions
- GroupBy
- OrderBy
- Limit (Interface only)
- API updates:
  - GroupBy/Expression
  - OrderBy/Ordering
  - Joins/Join
- Query code refactoring

### DIFF
--- a/shared/src/main/java/com/couchbase/lite/Function.java
+++ b/shared/src/main/java/com/couchbase/lite/Function.java
@@ -1,0 +1,44 @@
+package com.couchbase.lite;
+
+import java.util.Arrays;
+
+public class Function extends Expression {
+
+    private String func = null;
+    private Object param = null;
+
+    private Function(String func, Object param) {
+        this.func = func;
+        this.param = param;
+    }
+
+    public static Function avg(Object expr) {
+        return new Function("AVG()", expr);
+    }
+
+    public static Function count(Object expr) {
+        return new Function("COUNT()", expr);
+    } // null expression -> count *
+
+    public static Function min(Object expr) {
+        return new Function("MIN()", expr);
+    }
+
+    public static Function max(Object expr) {
+        return new Function("MAX()", expr);
+    }
+
+    public static Function sum(Object expr) {
+        return new Function("SUM()", expr);
+    }
+
+    @Override
+    protected Object asJSON() {
+        Object p = null;
+        if (param != null && param instanceof Expression)
+            p = ((Expression) param).asJSON();
+        else
+            p = param;
+        return Arrays.asList(func, p);
+    }
+}
--- a/shared/src/main/java/com/couchbase/lite/GroupBy.java
+++ b/shared/src/main/java/com/couchbase/lite/GroupBy.java
@@ -1,0 +1,48 @@
+package com.couchbase.lite;
+
+public class GroupBy extends Query implements HavingRouter, OrderByRouter, LimitRouter {
+
+    private Expression expr = null;
+
+    private GroupBy(Expression expr) {
+        this.expr = expr;
+    }
+
+    public static GroupBy expression(Expression expr) {
+        return new GroupBy(expr);
+    }
+
+    //---------------------------------------------
+    // implementation of HavingRouter
+    //---------------------------------------------
+    @Override
+    public Having having(Expression expr) {
+        return null;
+    }
+
+    //---------------------------------------------
+    // implementation of OrderByRouter
+    //---------------------------------------------
+    @Override
+    public OrderBy orderBy(OrderBy... orderBy) {
+        return null;
+    }
+
+    //---------------------------------------------
+    // implementation of LimitRouter
+    //---------------------------------------------
+
+    @Override
+    public Limit limit(Object limit) {
+        return null;
+    }
+
+    @Override
+    public Limit limit(Object limit, Object offset) {
+        return null;
+    }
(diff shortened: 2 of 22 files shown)


shared/src/main/java/com/couchbase/lite/Join.java
```diff
@@ -13,12 +13,10 @@
  */
 package com.couchbase.lite;
 
-import java.util.Arrays;
 import java.util.HashMap;
-import java.util.List;
 import java.util.Map;
 
-public class Join extends Query implements WhereRouter, OrderByRouter {
+public abstract class Join extends Query {
```
pasin: Join shouldn't extends Query. There is Joins which extends Query and implements WhereRouter and OrderByRouter.

shared/src/main/java/com/couchbase/lite/Query.java
```diff
@@ -211,7 +217,7 @@ protected void copy(Query query) {
     private void check() throws CouchbaseLiteException, IllegalStateException {
         database = (Database) from.getSource();
         String json = encodeAsJSON();
-        Log.v(TAG, "Query encoded as %s", json);
+        Log.e(TAG, "Query encoded as %s", json);
```
pasin: should be Log.v

shared/src/main/java/com/couchbase/lite/WhereRouter.java
```diff
@@ -14,6 +14,6 @@
 
 package com.couchbase.lite;
 
-/* package */  interface WhereRouter {
-    Where where(Expression expression);
+interface WhereRouter {
+    Where where(Expression expr);
```
pasin: I think it's better to make the variable name inline with the spec. Otherwise, each platform will have something totally different (I means API Doc / Guide). If you prefer short version, please create an issue to the spec and we will fix the spec and code at the same time. 

FYI, I have to go back and fix iOS code several times when I found that I don't have the name according to the spec.

hideki:
@pasin Thanks for reviewing PR.  I fixed the PR. Could you please review the PR again?

shared/src/main/java/com/couchbase/lite/Function.java
```diff
@@ -0,0 +1,44 @@
...
+
+    private String func = null;
+    private Object param = null;
+
+    private Function(String func, Object param) {
+        this.func = func;
+        this.param = param;
+    }
+
+    public static Function avg(Object expr) {
```
pasin: expression instead of expr per spec. This will be consistent with https://github.com/couchbase/couchbase-lite-android/blob/feature/2.0/shared/src/main/java/com/couchbase/lite/Expression.java as well.

shared/src/main/java/com/couchbase/lite/Function.java
```diff
@@ -0,0 +1,44 @@
...
+        return new Function("AVG()", expr);
+    }
+
+    public static Function count(Object expr) {
```
pasin: expression instead of expr per spec.

shared/src/main/java/com/couchbase/lite/Function.java
```diff
@@ -0,0 +1,44 @@
...
+        return new Function("COUNT()", expr);
+    } // null expression -> count *
+
+    public static Function min(Object expr) {
```
pasin: expression instead of expr per spec.

shared/src/main/java/com/couchbase/lite/Function.java
```diff
@@ -0,0 +1,44 @@
...
+        return new Function("MIN()", expr);
+    }
+
+    public static Function max(Object expr) {
```
pasin: expression instead of expr per spec.

shared/src/main/java/com/couchbase/lite/Function.java
```diff
@@ -0,0 +1,44 @@
...
+        return new Function("MAX()", expr);
+    }
+
+    public static Function sum(Object expr) {
```
pasin: expression instead of expr per spec.

shared/src/main/java/com/couchbase/lite/GroupBy.java
```diff
@@ -0,0 +1,68 @@
...
+        copy(query);
+        this.expressions = expressions;
+        setGroupBy(this);
+    }
+
+    //---------------------------------------------
+    // implementation of HavingRouter
+    //---------------------------------------------
+    @Override
+    public Having having(Expression expr) {
```
pasin: expression instead of expr.

pasin:
@hideki There are some more expr <> expression. I'm merging the PR now. Could you please review those and make changes later?

hideki:
@pasin Thanks for reviewing this. I will rename all `expr` to `expression`.

PR #1276, opened 2017-07-13
Fixed #1265 - 2.0: CouchbaseLiteException should be RuntimeException …

### LINKED ISSUE #1265: 2.0: CouchbaseLiteException should be RuntimeException or Exception
Currently CouchbaseLiteException extends from RuntimeException. However, some Exceptions could be thrown because of expected error like 409 - Conflict error. In some cases, CouchbaseLiteException is better to be extended from Exception.

Possible choices.
- Keep extending from RuntimeException
- Switch to extend from Exception.
- Define two CouchbaseLiteException(s) one is from RuntimeException, another one is from Exception.

hideki:
…or Exception

- Make CouchbaseLiteException extended from Exception instead of Runtime.
- Introduced CouchbaseLiteRuntimeException which could be thrown by inappropriate  codes
- API spec which has `thrown` keywords throws CouchbaseLiteException, others could throw CouchbaseLiteRuntimeException.

- In addition, removed /* package */ comment.
- some code formating.

### DIFF
--- a/shared/src/main/java/com/couchbase/lite/Database.java
+++ b/shared/src/main/java/com/couchbase/lite/Database.java
@@ -188,7 +188,7 @@ public final class Database {
      *
      * @param document
      */
-    public void save(Document document) {
+    public void save(Document document) throws CouchbaseLiteException {
         prepareDocument(document);
         document.save();
     }
@@ -203,7 +203,7 @@ public final class Database {
      *
      * @param document
      */
-    public void delete(Document document) {
+    public void delete(Document document) throws CouchbaseLiteException {
         prepareDocument(document);
         document.delete();
     }
@@ -214,7 +214,7 @@ public final class Database {
      *
      * @param document
      */
-    public void purge(Document document) {
+    public void purge(Document document) throws CouchbaseLiteException {
         prepareDocument(document);
         document.purge();
     }
@@ -229,7 +229,7 @@ public final class Database {
      * @param action the action which is implementation of Runnable interface
      * @throws CouchbaseLiteException Throws an exception if any error occurs during the operation.
      */
-    public void inBatch(Runnable action) throws CouchbaseLiteException, IllegalStateException {
+    public void inBatch(Runnable action) throws CouchbaseLiteException {
         mustBeOpen();
 
         try {
@@ -241,8 +241,6 @@ public final class Database {
                     action.run();
                     commit = true;
                 } catch (RuntimeException e) {
-                    if (e instanceof CouchbaseLiteException)
-                        throw e;
                     throw new CouchbaseLiteException(e);
                 }
             } finally {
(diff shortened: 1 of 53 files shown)


shared/src/main/java/com/couchbase/lite/Database.java
```diff
@@ -489,27 +487,26 @@ protected void finalize() throws Throwable {
...
+    Database copy() throws CouchbaseLiteException {
         return new Database(this.name, this.config);
     }
 
     //////// DATABASES:
 
     void mustBeOpen() {
         if (c4db == null)
-            throw new IllegalStateException("Database is not open.");
+            throw new CouchbaseLiteRuntimeException("Database is not open.");
```
pasin: I think we should throw IllegalStateException instead of CouchbaseLiteRuntimeException as it's self described that this is in an inconsistent or illegal state that users shouldn't do. CouchbaseLiteRuntimeException should be something that LC might thrown but not expected?

shared/src/main/java/com/couchbase/lite/ReadOnlyDocument.java
```diff
@@ -161,26 +164,23 @@ else if (c == '-')
...
 
-    /*package*/ ConflictResolver effectiveConflictResolver() {
+    ConflictResolver effectiveConflictResolver() {
         return getDatabase().getConflictResolver() != null ?
                 getDatabase().getConflictResolver() :
                 new DefaultConflictResolver();
     }
 
-    /* package */ boolean selectConflictingRevision() {
+    boolean selectConflictingRevision() throws CouchbaseLiteException {
```
pasin: Is CouchbaseLiteException thrown from anything in this method? 
Or it should return void but throw CouchbaseLiteException instead?

shared/src/main/java/com/couchbase/lite/ReadOnlyDocument.java
```diff
@@ -192,19 +192,20 @@ else if (c == '-')
         return true;
     }
 
-    /* package */ boolean selectCommonAncestor(ReadOnlyDocument doc1, ReadOnlyDocument doc2) {
+    boolean selectCommonAncestor(ReadOnlyDocument doc1, ReadOnlyDocument doc2)
+            throws CouchbaseLiteException {
```
pasin: Is CouchbaseLiteException thrown from anything in this method? 
Or it should return void but throw CouchbaseLiteException instead?

pasin:
As it has too many unrelated changes, I don't know if I can cover everything.

pasin:
For further PR, PLEASE separate unrelated stuff such as code formatting into separated commits. It's really difficult to review.

hideki:
@pasin, Sorry for including formatting changes in this PR.

hideki:
@pasin Applied your feedbacks to this PR. please review changes?

https://github.com/couchbase/couchbase-lite-android/pull/1276/commits/df8bd973e8005e147f24e056884df6496c9f5b25

PR #1303, opened 2017-07-26
DB13 examples

jamiltz:
@hideki could you review those changes? Not all changes are DB13 specific.


jamiltz:
@rajagp among other things, this PR tries to add an example of using a custom conflict resolver ([here](https://github.com/couchbase/couchbase-lite-android/pull/1303/files#diff-578cf04a86d519f929898b641f710f38)). To [create](https://github.com/couchbase/couchbase-lite-android/pull/1303/files#diff-a58e210fd99b3909fb87e0810ef23a1fR190) the conflict, I've been trying to create a revision on SG and on CBL before the replication starts but that's a bit hard to test. Have you managed to test your conflict resolver example by creating 2 local conflicts or is it also requiring SG to create the conflict?

rajagp:
@jamiltz : No SG involved - As you noted It's hard to get right . So just created 2 local conflicts. Simplest is to create two new documents with same docId. That should trigger the conflict resolver.

jamiltz:
@rajagp Ok I'll try that but in that case what is the `mine` and `theirs` property since both were created locally? I thought `theirs` would always be the one pulled from Sync Gateway but that doesn't happen with local conflicts.

rajagp:
@jamiltz  theirs is current rev in DB and mine is what's being saved.

jamiltz:
@rajagp Thanks that clears things up. Updated that example in the last commit. Unfortunately, it's still returning `mine` when reading back the document even though the conflict resolver is returning `conflict.getTheirs()`. I'd expect the `doc.status` to be `theirs`. Any ideas? (cc @hideki)

jamiltz:
Here's the screenshot of logcat https://cl.ly/1S203a240S0U/Screen%20Shot%202017-07-26%20at%2015.48.40.png

To test locally you can just checkout on that branch and build in Android Studio.

docs/examples/api-walkthrough/app/src/main/java/com/couchbase/apiwalkthrough/MainActivity.java
```diff
@@ -41,36 +43,59 @@ protected void onCreate(Bundle savedInstanceState) {
 
         // create database
         DatabaseConfiguration config = new DatabaseConfiguration(getApplicationContext());
-        final Database database = new Database("my-database", config);
+        config.setConflictResolver(new LocalWins());
+        Database database = null;
+        try {
+            database = new Database("my-database", config);
+        } catch (CouchbaseLiteException e) {
+            e.printStackTrace();
```
hideki: In case of Exception is thrown, the following code which refers database will cause NPE. Better to return here.

docs/examples/api-walkthrough/app/src/main/java/com/couchbase/apiwalkthrough/MainActivity.java
```diff
@@ -41,36 +43,59 @@ protected void onCreate(Bundle savedInstanceState) {
...
         // create document
         Document newTask = new Document();
         newTask.set("type", "task");
         newTask.set("owner", "todo");
         newTask.set("createdAt", new Date());
-        database.save(newTask);
+        try {
+            database.save(newTask);
+        } catch (CouchbaseLiteException e) {
+            e.printStackTrace();
```
hideki: Need some error handling here?

docs/examples/api-walkthrough/app/src/main/java/com/couchbase/apiwalkthrough/MainActivity.java
```diff
@@ -41,36 +43,59 @@ protected void onCreate(Bundle savedInstanceState) {
...
+        }
 
         // mutate document
         newTask.set("name", "Apples");
-        database.save(newTask);
+        try {
+            database.save(newTask);
+        } catch (CouchbaseLiteException e) {
+            e.printStackTrace();
```
hideki: same here

docs/examples/api-walkthrough/app/src/main/java/com/couchbase/apiwalkthrough/MainActivity.java
```diff
@@ -41,36 +43,59 @@ protected void onCreate(Bundle savedInstanceState) {
...
-            public void run() {
-                for (int i = 0; i < 10; i++) {
-                    Document doc = new Document();
-                    doc.set("type", "user");
-                    doc.set("name", String.format("user %s", i));
-                    database.save(doc);
-                    Log.d("app", String.format("saved user document %s", doc.getString("name")));
+        try {
+            final Database finalDatabase = database;
+            database.inBatch(new TimerTask() {
```
hideki: I think better to use `Runnable` instead of `TimerTask`.

jamiltz:
@rajagp Made some progress. It returns `theirs` if I delete the app from the emulator first and then run it. After updating the conflict resolver to `return mine;`, it logs `mine` as expected. Then if I update it back to `return theirs;`, it still logs `mine` instead of `theirs`. So it looks to be related to what is already in the database. Same result when running the code you suggested or the one currently on this branch. Sounds like it's partly working as expected now so I'll put that code snippet in the docs. Thanks

docs/examples/api-walkthrough/app/src/main/java/com/couchbase/apiwalkthrough/MainActivity.java
jamiltz: done.

docs/examples/api-walkthrough/app/src/main/java/com/couchbase/apiwalkthrough/MainActivity.java
jamiltz: done.

rajagp:
@jamiltz ...little confusing. did you file a ticket  ?

jamiltz:
@hideki updated per your review and rebased in a single commit. LGTM

PR #1318, opened 2017-07-28
Fixed #1201 - Thread safety for Database instance.

### LINKED ISSUE #1201: 2.0: Thread Safe (Database)

hideki:
- Database class has a lock object for synchronization.
- Synchronization should be done in public method level.
- In case DB is not open, throw CouchbaseLiteRuntimeException instead of IllegalStateException. To be consistent with case db close state problem is found in the LiteCore level.

### DIFF
--- a/android/CouchbaseLite/src/androidTest/assets/test.properties
+++ b/android/CouchbaseLite/src/androidTest/assets/test.properties
@@ -1,5 +1,6 @@
 # properties file for unit test
 replicatorTestsEnabled=false
+threadTestsEnabled=false
 remoteHost=10.0.2.2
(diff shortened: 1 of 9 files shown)


pasin:
I understand that you need to lock resolveConflictInDocument(), postDatabaseChanged(), and postDocumentChanged() as well. 

When locking postDatabaseChanged() and postDocumentChanged(), calling listener.change(change) needs to be outside the synchronized block.

hideki:
@pasin,
I will track this following ticket.
https://github.com/couchbase/couchbase-lite-android/issues/1321

PR #1514, opened 2018-01-14
Fixed #1513 - 2.0: API updates 01/12/2018

### LINKED ISSUE #1513: 2.0: API updates 01/12/2018
Another API updates

https://github.com/couchbaselabs/couchbase-lite-apiv2/commit/17133caeb0b3a3d441f2e25d62e3d4d570261cad
https://github.com/couchbaselabs/couchbase-lite-apiv2/commit/ec1769fd07e6f3b3eee4b9662465828ae1b2d640
https://github.com/couchbaselabs/couchbase-lite-apiv2/commit/552500a7e50297eb504864842a6094a64332c16e

### ResultSet
- [x] Change from ResultSet.toArray() to ResultSet.allResults().
- [x] allResults() -> Array<Result> method will return the current unenumerated Result objects. After calling this method, all of the Result objects will be enumerated. Subsequence calls to next() method will return null. Subsequence calls to allResults() will return an empty array.
  - [x] Why? The goal here is to make ResultSet API forward enumeration only at least for now. When LiteCore has done the optimization, we will revisit this again.
  - [x] For iOS, this is the same behavior as the spec of the NSEnumerator’s -allObject method.
  - [x] @Jim and @hideki. Please let me know if this works for .NET and Java regarding implementation and platform specific conflicts.

### Expression
- [x] Remove concat() for 2.0. We should do this in 2.x.
- [x] Fixed all() method to return PropertyExpression instead of FromExpression. Hence, remove FromExpression.
- [x] Fixed ArrayExpression.any(variable), .every(variable), anyAndEvery(variable) to take VariableExpression instead of String name. Hence, Expression.variable(name) returns VariableExpression now.
- [x] Please make sure to have isNOT(expression). I sent out an update about this before, but I have missed updating it in the spec.

### Type Safe Expression API
- [x] We decided to make the Expression API type safe. 
  - [x] For example, change Expression.multiply(Object) to Expression.multiply(Expression) or Function.count(Object) to Function.count(Expression).
- [x] Add value expressions such as Expression.value(value), Expression.string(value), and so on.
  - [x] We will not support Blob, Dictionary, and Array value type in 2.0. We could think about adding these based on demands later.
  - [x] For the languages that support Literal Convertibles such as Swift, we could later add Literal Convertibles to automatically convert literal values to Expressions as well in 2.x.

### Database
- [x] (Revised Behavior) delete(doc) :
  - [x] Throws not-allow exception when deleting a newly created doc (no RevID).
  - [x] No-ops when deleting a non existing document.
  - [x] Go to conflict handler when delete an already deleted document. This will make it the same behavior as when pulling deleted documents. Also the default conflict resolver will handle this by returning theirs version which will be equal to no-ops.
- [x] (Revised Behavior) purge(doc):
  - [x] Throws not-allow exception when purging a newly created doc (no RevID).
  - [x] No-ops when purging a non existing document. For implementation, I understand LC has already make the operation no-ops.

hideki:
- Please refer ticket, #1513,   for items


shared/src/main/java/com/couchbase/lite/ResultSet.java
```diff
@@ -84,30 +87,33 @@ public Result next() {
...
+    /**
+     * Return List of Results. List is unmodifiable and only supports
+     * int get(int index), int size(), boolean isEmpty() and Iterator<Result> iterator() methods.
+     * Once called allResults(), next() method return null. Don't call next() and allResults()
+     * together.
+     *
+     * @return List of Results
+     */
+    public List<Result> allResults() {
+        randomAccess = true;
```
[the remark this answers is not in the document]
pasin: To have the same behavior as the spec, can we just iterate through all of the result and return the list?

I know that the current implementation will have a better performance but it needs to call c4enum.getRowCount() and c4enum.seek(index) which may or may not exist after LC optimizes the c4enum. You could keep the implementation of some of the methods for future use.

shared/src/main/java/com/couchbase/lite/ResultSet.java
```diff
@@ -70,10 +71,12 @@ public Result next() {
         if (query == null)
             throw new IllegalStateException("_query variable is null");
 
+        if (randomAccess)
```
pasin: randomAccess is not thread safe.

shared/src/main/java/com/couchbase/lite/Expression.java
```diff
@@ -342,48 +426,54 @@ public String toString() {
...
-            return value;
-    }
-
     abstract Object asJSON();
 
+    static final class ValueExpression extends Expression {
+        private Object value;
+
+        ValueExpression(Object value) {
+            this.value = value;
```
pasin: We need to validate the type of the Object here. Currently we support String, Number, int, long, float, double, boolean, Date, and null.

shared/src/main/java/com/couchbase/lite/Expression.java
```diff
@@ -342,48 +426,54 @@ public String toString() {
...
+        }
+
+        @Override
+        Object asJSON() {
+            return this.value;
```
pasin: If the value is Date, I understand that we need to convert to JSON date string.

hideki: Thanks for this.

hideki:
Notes to self:

Because of the following `C4QueryEnumerator` implementation, `List<Result> ResultSet.allResults()` can not behave as we expected.

```c++
    /** A query result enumerator.
        Created by c4db_query. Must be freed with c4queryenum_free.
        The fields of this struct represent the current matched index row, and are valid until the
        next call to c4queryenum_next or c4queryenum_free. */
    typedef struct {
        /** The columns of this result, in the same order as in the query's `WHAT` clause. */
        FLArrayIterator columns;

        /** A bitmap where a 1 bit represents a column whose value is MISSING.
            This is how you tell a missing property value from a value that's JSON 'null',
            since the value in the `columns` array will be a Fleece `null` either way. */
        uint64_t missingColumns;

        /** The number of full-text matches (i.e. the number of items in `fullTextMatches`) */
        uint32_t fullTextMatchCount;

        /** Array with details of each full-text match */
        const C4FullTextMatch *fullTextMatches;
    } C4QueryEnumerator;
```

hideki:
Notes to Self:
If we need to avoid to use  c4enum.getRowCount() and c4enum.seek(index), It is difficult to implemt `ResultSet.iterator()` too.

hideki:
@pasin 
Fixed the conflict. If PR is fine, please merge. 
Thanks!

PR #1789, opened 2018-12-04
Document Replication Eventing

Sandychuang8:
- add replicator progress level enum, document replicated status class, change name from document error to document ended and fire replicated ended when document replicated.

- add document replicated listener, and listener token, document replicated update class setter/getter progress level, allow user to add/remove replication listener.

- add unit test to test push/pull document replicated.

### DIFF
--- a/android/CouchbaseLite/src/androidTest/java/com/couchbase/lite/ReplicatorWithSyncGatewayDBTest.java
+++ b/android/CouchbaseLite/src/androidTest/java/com/couchbase/lite/ReplicatorWithSyncGatewayDBTest.java
@@ -31,6 +31,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.net.URI;
 import java.net.URISyntaxException;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Locale;
@@ -58,9 +59,11 @@ import static org.junit.Assert.fail;
  * configuration file.
  */
 public class ReplicatorWithSyncGatewayDBTest extends BaseReplicatorTest {
+
     public static final MediaType JSON = MediaType.parse("application/json; charset=utf-8");
 
     private static final String DB_NAME = "db";
+    private List<DocumentReplicatedUpdate> replicationEvents = new ArrayList<DocumentReplicatedUpdate>();
 
     @Before
     public void setUp() throws Exception {
@@ -609,5 +612,80 @@ public class ReplicatorWithSyncGatewayDBTest extends BaseReplicatorTest {
 
         // "because the replicator was reset"
         assertEquals(2L, db.getCount());
+    }
+
+    @Test
+    public void testReplicationListenerForDocumentEndedStatus() throws URISyntaxException, InterruptedException, CouchbaseLiteException {
+        if (!config.replicatorTestsEnabled())
+            return;
+
+        boolean ret = false;
+        MutableDocument doc1 = new MutableDocument("doc1");
+        doc1.setString("name", "Tiger");
+        db.save(doc1);
+
+        // target SG URI
+        Endpoint target = getRemoteEndpoint(DB_NAME, false);
+
+        // Push replicate from db to SG
+        ReplicatorConfiguration config = makeConfig(true, true, true, this.otherDB, target);//push n pull
+
+        repl = new Replicator(config);
+        final CountDownLatch latch = new CountDownLatch(1);
+        ListenerToken replicationListener = repl.addReplicationListener(executor, new DocumentReplicatedListener() {
+            @Override
+            public void replicated(DocumentReplicatedUpdate update) {
+                replicationEvents.add(update);
+            }
+        });
+
+        ListenerToken token = repl.addChangeListener(executor, new ReplicatorChangeListener() {
+            @Override
+            public void changed(ReplicatorChange change) {
+                Replicator.Status status = change.getStatus();
+                CouchbaseLiteException error = status.getError();
+                final String kActivityNames[] = {"stopped", "offline", "connecting", "idle", "busy"};
+                Log.i(TAG, "ReplicatorChangeListener.changed() status: %s (%d / %d), lastError = %s",
+                        kActivityNames[status.getActivityLevel().getValue()],
+                        status.getProgress().getCompleted(), status.getProgress().getTotal(),
+                        error);
+                //isContinuous
+                if (status.getActivityLevel() == Replicator.ActivityLevel.IDLE &&
+                        status.getProgress().getCompleted() == status.getProgress().getTotal()) {
+
+                    assertNull(error);
+
+                    latch.countDown();
+                }
+            }
+
+        });
+
+        repl.start();
+
+        MutableDocument doc2 = new MutableDocument("doc2");
+        doc2.setString("name", "Cat");
+        otherDB.save(doc2);
+
+        try {
+           ret  = latch.await(30, TimeUnit.SECONDS); //wait 30 sec
+        } catch (Exception e){
+            repl.stop();
+            throw e;
+        } finally{
+            assertTrue(ret);
+            repl.removeChangeListener(token);
+            repl.removeChangeListener(replicationListener);
+        }
+
+        assertEquals(replicationEvents.size(),2);
+        for (DocumentReplicatedUpdate d:replicationEvents
+             ) {
+            if(d.getStatus().getPushingFlag() == true && d.getStatus().getCompletedFlag() == true){
+                assertEquals(d.getStatus().getDocId(),"doc1");
+            } else if(d.getStatus().getCompletedFlag() == false && d.getStatus().getCompletedFlag() == true){
+                assertEquals(d.getStatus().getDocId(),"doc2");
+            }
+        }
     }
 }
--- a/android/CouchbaseLite/src/androidTest/java/com/couchbase/lite/ReplicatorWithSyncGatewayDBTest.java
+++ b/android/CouchbaseLite/src/androidTest/java/com/couchbase/lite/ReplicatorWithSyncGatewayDBTest.java
@@ -31,7 +31,6 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.net.URI;
 import java.net.URISyntaxException;
-import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Locale;
--- a/android/CouchbaseLite/src/androidTest/java/com/couchbase/lite/ReplicatorWithSyncGatewayDBTest.java
+++ b/android/CouchbaseLite/src/androidTest/java/com/couchbase/lite/ReplicatorWithSyncGatewayDBTest.java
@@ -63,7 +63,6 @@ public class ReplicatorWithSyncGatewayDBTest extends BaseReplicatorTest {
     public static final MediaType JSON = MediaType.parse("application/json; charset=utf-8");
 
     private static final String DB_NAME = "db";
-    private List<DocumentReplicatedUpdate> replicationEvents = new ArrayList<DocumentReplicatedUpdate>();
 
     @Before
     public void setUp() throws Exception {
@@ -612,80 +611,5 @@ public class ReplicatorWithSyncGatewayDBTest extends BaseReplicatorTest {
 
         // "because the replicator was reset"
         assertEquals(2L, db.getCount());
-    }
-
-    @Test
-    public void testReplicationListenerForDocumentEndedStatus() throws URISyntaxException, InterruptedException, CouchbaseLiteException {
-        if (!config.replicatorTestsEnabled())
-            return;
-
-        boolean ret = false;
-        MutableDocument doc1 = new MutableDocument("doc1");
-        doc1.setString("name", "Tiger");
-        db.save(doc1);
-
-        // target SG URI
-        Endpoint target = getRemoteEndpoint(DB_NAME, false);
-
-        // Push replicate from db to SG
-        ReplicatorConfiguration config = makeConfig(true, true, true, this.otherDB, target);//push n pull
-
-        repl = new Replicator(config);
-        final CountDownLatch latch = new CountDownLatch(1);
-        ListenerToken replicationListener = repl.addReplicationListener(executor, new DocumentReplicatedListener() {
-            @Override
-            public void replicated(DocumentReplicatedUpdate update) {
-                replicationEvents.add(update);
-            }
-        });
-
-        ListenerToken token = repl.addChangeListener(executor, new ReplicatorChangeListener() {
-            @Override
-            public void changed(ReplicatorChange change) {
-                Replicator.Status status = change.getStatus();
-                CouchbaseLiteException error = status.getError();
-                final String kActivityNames[] = {"stopped", "offline", "connecting", "idle", "busy"};
-                Log.i(TAG, "ReplicatorChangeListener.changed() status: %s (%d / %d), lastError = %s",
-                        kActivityNames[status.getActivityLevel().getValue()],
-                        status.getProgress().getCompleted(), status.getProgress().getTotal(),
-                        error);
-                //isContinuous
-                if (status.getActivityLevel() == Replicator.ActivityLevel.IDLE &&
-                        status.getProgress().getCompleted() == status.getProgress().getTotal()) {
-
-                    assertNull(error);
-
-                    latch.countDown();
-                }
-            }
-
-        });
-
-        repl.start();
-
-        MutableDocument doc2 = new MutableDocument("doc2");
-        doc2.setString("name", "Cat");
-        otherDB.save(doc2);
-
-        try {
-           ret  = latch.await(30, TimeUnit.SECONDS); //wait 30 sec
-        } catch (Exception e){
-            repl.stop();
-            throw e;
-        } finally{
-            assertTrue(ret);
-            repl.removeChangeListener(token);
-            repl.removeChangeListener(replicationListener);
-        }
-
-        assertEquals(replicationEvents.size(),2);
-        for (DocumentReplicatedUpdate d:replicationEvents
-             ) {
-            if(d.getStatus().getPushingFlag() == true && d.getStatus().getCompletedFlag() == true){
-                assertEquals(d.getStatus().getDocId(),"doc1");
-            } else if(d.getStatus().getCompletedFlag() == false && d.getStatus().getCompletedFlag() == true){
-                assertEquals(d.getStatus().getDocId(),"doc2");
-            }
-        }
     }
 }
--- a/android/CouchbaseLite/src/ce/java/com/couchbase/lite/ReplicatorConfiguration.java
+++ b/android/CouchbaseLite/src/ce/java/com/couchbase/lite/ReplicatorConfiguration.java
@@ -43,6 +43,7 @@ public final class ReplicatorConfiguration {
     static final String kC4ReplicatorCheckpointInterval = "checkpointInterval"; // How often to checkpoint, in seconds; number
     static final String kC4ReplicatorOptionRemoteDBUniqueID = "remoteDBUniqueID"; // How often to checkpoint, in seconds; number
     static final String kC4ReplicatorResetCheckpoint = "reset"; // reset remote checkpoint
+    static final String kC4ReplicatorOptionProgressLevel = "progress";  //< If >=1, notify on every doc; if >=2, on every attachment (int)
 
     // Auth dictionary keys:
     static final String kC4ReplicatorAuthType = "type"; // Auth property; string
@@ -94,6 +95,7 @@ public final class ReplicatorConfiguration {
     private byte[] pinnedServerCertificate = null;
     private List<String> channels = null;
     private List<String> documentIDs = null;
+    private ReplicatorProgressLevel progressLevel = ReplicatorProgressLevel.OVERALL;
 
     //---------------------------------------------
     // Constructors
@@ -110,6 +112,7 @@ public final class ReplicatorConfiguration {
         this.headers = config.headers;
         this.channels = config.channels;
         this.documentIDs = config.documentIDs;
+        this.progressLevel = config.progressLevel;
     }
 
     public ReplicatorConfiguration(Database database, Endpoint target) {
@@ -224,6 +227,13 @@ public final class ReplicatorConfiguration {
         return this;
     }
 
+    public ReplicatorConfiguration setProgressLevel(ReplicatorProgressLevel level){
+        if(readonly)
+            throw new IllegalStateException("ReplicatorConfiguration is readonly mode.");
+        this.progressLevel = level;
+        return this;
+    }
+
     //---------------------------------------------
     // Getters
     //---------------------------------------------
@@ -294,6 +304,8 @@ public final class ReplicatorConfiguration {
     public List<String> getDocumentIDs() {
         return documentIDs;
     }
+
+    public ReplicatorProgressLevel getProgressLevel() { return  progressLevel; }
 
     //---------------------------------------------
     // Package level access
--- a/android/CouchbaseLite/src/ce/java/com/couchbase/lite/ReplicatorConfiguration.java
+++ b/android/CouchbaseLite/src/ce/java/com/couchbase/lite/ReplicatorConfiguration.java
@@ -69,6 +69,18 @@ public final class ReplicatorConfiguration {
         PULL
     }
 
+    /**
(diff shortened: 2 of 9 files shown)


android/CouchbaseLite/src/ce/java/com/couchbase/lite/ReplicatorConfiguration.java
```diff
@@ -82,6 +95,7 @@
     private byte[] pinnedServerCertificate = null;
     private List<String> channels = null;
     private List<String> documentIDs = null;
+    private ReplicatorProgressLevel progressLevel = ReplicatorProgressLevel.OVERALL;
```
pasin: This should belong to AbstractReplicator not ReplicatorConfiguration.

android/CouchbaseLite/src/ce/java/com/couchbase/lite/ReplicatorConfiguration.java
```diff
@@ -212,6 +227,13 @@ public ReplicatorConfiguration setDocumentIDs(List<String> documentIDs) {
         return this;
     }
 
+    public ReplicatorConfiguration setProgressLevel(ReplicatorProgressLevel level){
```
pasin: We don't allow users to set the progress level. Please remove from the API.

android/CouchbaseLite/src/ce/java/com/couchbase/lite/ReplicatorConfiguration.java
```diff
@@ -283,6 +305,8 @@ public Authenticator getAuthenticator() {
         return documentIDs;
     }
 
+    public ReplicatorProgressLevel getProgressLevel() { return  progressLevel; }
```
pasin: We don't allow user to set/get the progress level. Please remove this.

android/CouchbaseLite/src/ce/java/com/couchbase/lite/ReplicatorConfiguration.java
```diff
@@ -212,6 +227,13 @@ public ReplicatorConfiguration setDocumentIDs(List<String> documentIDs) {
         return this;
     }
 
+    public ReplicatorConfiguration setProgressLevel(ReplicatorProgressLevel level){
+        if(readonly)
+            throw new IllegalStateException("ReplicatorConfiguration is readonly mode.");
+        this.progressLevel = level;
```
pasin: This doesn't belong here. Please remove.

shared/src/main/java/com/couchbase/lite/AbstractReplicator.java
```diff
@@ -224,6 +225,65 @@ Status copy() {
         }
     }
 
+    /**
+     * Document ended progress of a replicator.
+     */
+    public final static class DocumentReplicatedStatus {
```
pasin: 1. The correct name is DocumentReplication.
2. The correct API is here : https://github.com/couchbaselabs/couchbase-lite-apiv2/blob/master/spec/spec/Replicator/Replicator.swift#L67-L72.
3. Please extract the class to its own java file.

shared/src/main/java/com/couchbase/lite/AbstractReplicator.java
```diff
@@ -371,6 +433,44 @@ public void removeChangeListener(ListenerToken token) {
         }
     }
 
+    /**
+     * Set the given DocumentReplicatedListener to the this replicator.
+     *
+     * @param listener
+     * @return ListenerToken A token to remove the handler later
+     */
+    public ListenerToken addReplicationListener(DocumentReplicatedListener listener) {
```
pasin: The correct signature is `addDocumentReplicationListener(DocumentReplicationListener listener)`.

shared/src/main/java/com/couchbase/lite/AbstractReplicator.java
```diff
@@ -371,6 +433,44 @@ public void removeChangeListener(ListenerToken token) {
...
+        return addReplicationListener(null, listener);
+    }
+
+    /**
+     * Set the given DocumentReplicatedListener to the this replicator.
+     *
+     * @param listener
+     */
+    public ListenerToken addReplicationListener(Executor executor, DocumentReplicatedListener listener) {
```
pasin: The correct signature is `addDocumentReplicationListener(Executor executor, DocumentReplicationListener listener)`.

shared/src/main/java/com/couchbase/lite/AbstractReplicator.java
```diff
@@ -371,6 +433,44 @@ public void removeChangeListener(ListenerToken token) {
...
+            DocumentReplicatedListenerToken token = new DocumentReplicatedListenerToken(executor, listener);
+            docEndedListenerTokens.add(token);
+            return token;
+        }
+    }
+
+    /**
+     * Remove the given DocumentReplicatedListener from the this replicator.
+     */
+    public void removeReplicationListener(ListenerToken token) {
```
pasin: We don't have a specific removeReplicationListener(token). Please change removeChangeListener(token) to be able to remove DocumentReplicationListener as well.

shared/src/main/java/com/couchbase/lite/AbstractReplicator.java
```diff
@@ -371,6 +433,44 @@ public void removeChangeListener(ListenerToken token) {
...
+        synchronized (lock) {
+            if (token == null || !(token instanceof DocumentReplicatedListenerToken))
+                throw new IllegalArgumentException();
+            this.config.setProgressLevel(ReplicatorConfiguration.ReplicatorProgressLevel.OVERALL);
```
pasin: This logic is also not correct because there could be multiple `DocumentReplicationListener` added so need to set to `OVERALL` only when all of the `DocumentReplicationListener` objects added were removed.

shared/src/main/java/com/couchbase/lite/AbstractReplicator.java
```diff
@@ -513,13 +613,13 @@ public void run() {
             }
 
             @Override
-            public void documentError(C4Replicator repl, final boolean pushing, final String docID, final C4Error error, final boolean trans, Object context) {
+            public void documentEnded(C4Replicator repl, final boolean pushing, final String docID, final C4Error error, final boolean trans, Object context) {
```
pasin: This signature seems to be old. Please update LiteCore.

shared/src/main/java/com/couchbase/lite/AbstractReplicator.java
```diff
@@ -563,6 +663,13 @@ private void documentError(boolean pushing, String docID, C4Error error, boolean
             Log.i(TAG, "C4ReplicatorListener.documentError() pushing -> %s, docID -> %s, error -> %s, trans -> %s", pushing ? "true" : false, docID, error, trans ? "true" : false);
             // TODO: Should pass error along to listener
         }
+
+        if(error.getDomain()==0 && error.getCode()==0){
+            //DocumentReplicatedStatus status = new DocumentReplicatedStatus(true, pushing, docID);
+            c4ReplListener.documentEnded(c4repl, pushing, docID,
```
pasin: This seems to be incorrect. Instead we need to notify the listener added to the replicator.

shared/src/main/java/com/couchbase/lite/DocumentReplicatedListener.java
```diff
@@ -0,0 +1,10 @@
+package com.couchbase.lite;
+
+public interface DocumentReplicatedListener {
```
pasin: The correct name is `DocumentReplicationListener`.

shared/src/main/java/com/couchbase/lite/DocumentReplicatedListener.java
```diff
@@ -0,0 +1,10 @@
...
+    /**
+     * The callback function from Replicator
+     *
+     * @param update the Document replicated information
+     */
+    void replicated(DocumentReplicatedUpdate update);
```
pasin: The correct class is DocumentReplication.

shared/src/main/java/com/couchbase/lite/AbstractReplicator.java
pasin: After all, I think we don't need this class as there is another one below called `DocumentReplicatedUpdate`.

shared/src/main/java/com/couchbase/lite/DocumentReplicatedUpdate.java
```diff
@@ -0,0 +1,34 @@
+package com.couchbase.lite;
+
+public final class DocumentReplicatedUpdate {
```
pasin: 1. The correct name is DocumentReplication.
2. The correct API is here : https://github.com/couchbaselabs/couchbase-lite-apiv2/blob/master/spec/spec/Replicator/Replicator.swift#L67-L72.

shared/src/main/java/com/couchbase/lite/ReplicatorChangeListenerToken.java
```diff
@@ -43,3 +43,28 @@ Executor getExecutor() {
         return executor != null ? executor : DefaultExecutor.instance();
     }
 }
+
+final class DocumentReplicatedListenerToken implements ListenerToken {
```
pasin: Please change the name to DocumentReplicationListenerToken.

android/CouchbaseLite/src/androidTest/java/com/couchbase/lite/ReplicatorWithSyncGatewayDBTest.java
```diff
@@ -31,6 +31,7 @@
 import java.io.InputStream;
 import java.net.URI;
 import java.net.URISyntaxException;
+import java.util.ArrayList;
```
pasin: Not sure if this is needed.

shared/src/main/java/com/couchbase/lite/AbstractReplicator.java
Sandychuang8: DocumentReplicatedListener is listen for document replicated event, so shouldn't DocumentReplicatedListener be better?

shared/src/main/java/com/couchbase/lite/AbstractReplicator.java
```diff
@@ -687,17 +629,20 @@ private void documentEnded(boolean pushing, String docID, String revID, int flag
                 Log.e(TAG, "Failed to resolveConflict: docID -> %s", ex, docID);
                 // TODO: Should pass error along to listener
```
pasin: We need to send DocumentReplication as an error for this case as well.

shared/src/main/java/com/couchbase/lite/AbstractReplicator.java
```diff
@@ -687,17 +629,20 @@ private void documentEnded(boolean pushing, String docID, String revID, int flag
...
             }
-        } else {
+        } else if(error.getDomain()==0 && error.getCode()==0) {
```
pasin: We should notify when there is an error as well.

shared/src/main/java/com/couchbase/lite/DocumentReplicatedListener.java
pasin: The class name hasn't been correct yet. Could you change it to DocumentReplicationListener?

shared/src/main/java/com/couchbase/lite/DocumentReplicatedUpdate.java
pasin: The class name is still incorrect. Could you please cross-check with the API spec?

shared/src/main/java/com/couchbase/lite/DocumentReplicatedUpdate.java
```diff
@@ -0,0 +1,100 @@
...
+    private boolean isDeleted = false;
+    private boolean pushing = false;
+    private String docId = "";
+
+    //---------------------------------------------
+    // Constructors
+    //---------------------------------------------
+    DocumentReplicatedUpdate(Replicator replicator, boolean completed, boolean isDeleted, boolean pushing, String docId) {
+        this.replicator = replicator;
+        this.completed = completed;
```
pasin: We don't have completed in the API spec.

shared/src/main/java/com/couchbase/lite/ReplicatorChangeListenerToken.java
pasin: Could you change the name to DocumentReplicationListenerToken?

shared/src/main/java/com/couchbase/lite/AbstractReplicator.java
```diff
@@ -361,13 +399,41 @@ public ListenerToken addChangeListener(Executor executor, ReplicatorChangeListen
...
-     * Remove the given ReplicatorChangeListener from the this replicator.
+     * Remove the given ReplicatorChangeListener or DocumentReplicatedListener from the this replicator.
      */
     public void removeChangeListener(ListenerToken token) {
         synchronized (lock) {
-            if (token == null || !(token instanceof ReplicatorChangeListenerToken))
+            if (token == null || (!(token instanceof ReplicatorChangeListenerToken) && !(token instanceof DocumentReplicatedListenerToken)))
                 throw new IllegalArgumentException();
-            changeListenerTokens.remove(token);
+            if(docEndedListenerTokens.remove(token)==true) {
```
pasin: I think this should check the size instead of the result of remove as below.

```Java
docEndedListenerTokens.remove(token);
if (docEndedListenerTokens.size() == 0) {
    setProgressLevel(ReplicatorProgressLevel.OVERALL);
}
```

Sandychuang8: question, is setProgressLevel(ReplicatorProgressLevel.OVERALL); expensive? if (docEndedListenerTokens.size() == 0) check will be access many times even the doc Ended listener was never added...

pasin: I’m not sure I understand your point. It’s not about performance here. The reason to set the progress level to overall when the size of docEndedListenerToken set is zero because there could be more than one listener added; set to overall should be done when all listeners are removed.

shared/src/main/java/com/couchbase/lite/DocumentReplication.java
```diff
@@ -0,0 +1,133 @@
...
+     * The current document replication direction flag.
+     */
+    public boolean isPush() {
+        return pushing;
+    }
+
+    /**
+     * The current document id.
+     */
+    public String getDocumentId() {
```
pasin: should be getDocumentID() for consistency with the other parts of the API.

shared/src/main/java/com/couchbase/lite/DocumentReplication.java
```diff
@@ -0,0 +1,133 @@
...
+        return docId;
+    }
+
+    /**
+     * The current document revision id.
+     */
+    public String getRevisionId() {
```
pasin: Please remove. We don't have this in the API.

shared/src/main/java/com/couchbase/lite/DocumentReplication.java
```diff
@@ -0,0 +1,133 @@
...
+     * The current document id.
+     */
+    public boolean isDeleted() {
+        return isDeleted;
+    }
+
+    /**
+     * The current document flags.
+     */
+    public int getFlags() {
```
pasin: Please remove. We don't have this in the API.

shared/src/main/java/com/couchbase/lite/DocumentReplication.java
```diff
@@ -0,0 +1,133 @@
...
+     * The current document revision id.
+     */
+    public String getRevisionId() {
+        return revId;
+    }
+
+    /**
+     * The current document id.
+     */
+    public boolean isDeleted() {
```
pasin: Please remove for now. Most likely we may add it back.

shared/src/main/java/com/couchbase/lite/DocumentReplication.java
```diff
@@ -0,0 +1,133 @@
...
+     * The current document flags.
+     */
+    public int getFlags() {
+        return flags;
+    }
+
+    /**
+     * The current document replication error.
+     */
+    public C4Error getError() {
```
pasin: C4Error is private so return CouchbaseLiteException instead. To convert to CouchbaseLiteException, you can use  CBLStatus.convertError() method. Please make sure to to check handle null error as well.

shared/src/main/java/com/couchbase/lite/DocumentReplication.java
```diff
@@ -0,0 +1,133 @@
...
+                ", is pushing =" + pushing +
+                ", document id =" + docId +
+                ", revision id =" + revId +
+                ", error code =" + error.getCode()+
+                ", error domain=" + error.getDomain() +
+                ", is transient=" + trans +
+                ", flags =" + flags +
+                ", doc is deleted =" + isDeleted +
+                '}';
+    }
```
pasin: Please remove info that is not in public API.

shared/src/main/java/com/couchbase/lite/DocumentReplication.java
```diff
@@ -0,0 +1,133 @@
...
+     * The current document replication error.
+     */
+    public C4Error getError() {
+        return error;
+    }
+
+    /**
+     * The current document replication transient.
+     */
+    public boolean getTransient() {
```
pasin: Please remove. We don't have this in the API.

shared/src/main/java/com/couchbase/lite/AbstractReplicator.java
```diff
@@ -106,6 +109,38 @@ int getValue() {
...
+        private final int value;
+        private static Map map = new HashMap<>();
+
+        private ReplicatorProgressLevel(int value) {
+            this.value = value;
+        }
+        static {
+            for (ReplicatorProgressLevel level : ReplicatorProgressLevel.values()) {
+                map.put(level.value, level);
+            }
```
pasin: I think we don't need to put the enum value into the map.

shared/src/main/java/com/couchbase/lite/AbstractReplicator.java
```diff
@@ -106,6 +109,38 @@ int getValue() {
...
+        }
+
+        public static ReplicatorProgressLevel valueOf(int level) {
+            return (ReplicatorProgressLevel) map.get(level);
```
pasin: Are we using this somewhere? I couldn't find. If not, please remove.

shared/src/main/java/com/couchbase/lite/AbstractReplicator.java
```diff
@@ -106,6 +109,38 @@ int getValue() {
...
+     * PER_DOCUMENT: >=1 Every document replication ended callback
+     * PER_ATTACHMENT: >=2 Every blob replication progress callback
+     */
+    enum ReplicatorProgressLevel {
+        OVERALL(0),
+        PER_DOCUMENT(1),
+        PER_ATTACHMENT(2);
+
+        private final int value;
+        private static Map map = new HashMap<>();
```
pasin: I'm not sure if we need to store value in the map.

shared/src/main/java/com/couchbase/lite/DocumentReplicationListener.java
```diff
@@ -0,0 +1,28 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+package com.couchbase.lite;
+
+public interface DocumentReplicationListener {
```
pasin: Can you add a little bit of JAVADOC comment about DocumentReplicationListener?

shared/src/main/java/com/couchbase/lite/DocumentReplication.java
Sandychuang8: Base on https://github.com/couchbase/couchbase-lite-android/issues/1791, I thought we need to add isDeleted property?

pasin: Yes, I think we add that in a separate PR which should also include the test for it too.

Sandychuang8: yes, I have the tests for isDeleted already.

shared/src/main/java/com/couchbase/lite/DocumentReplication.java
```diff
@@ -72,62 +72,44 @@ public boolean isPush() {
...
+        return isAccessRemoved;
+    } //TODO: will find a way to get this value and write unit test for it in the related PR for the ticket.
 
     /**
      * The current document replication error.
      */
-    public C4Error getError() {
-        return error;
+    public CouchbaseLiteException getError() {
+        return error.getCode() != 0 ? CBLStatus.convertError(error) : null;
```
pasin: Just want to check if it is possible that the error could null itself? Could you please confirm?

shared/src/main/java/com/couchbase/lite/DocumentReplication.java
```diff
@@ -0,0 +1,115 @@
...
+    public boolean isDeleted() {
+        return isDeleted;
+    }
+
+    /**
+     * The current access removed status of the document.
+     */
+    public boolean isAccessRemoved()  {
+        return isAccessRemoved;
+    } //TODO: will find a way to get this value and write unit test for it in the related PR for the ticket.
```
pasin: Is this TODO: still valid?

shared/src/main/java/com/couchbase/lite/DocumentReplication.java
```diff
@@ -0,0 +1,115 @@
...
+
+    @Override
+    public String toString() {
+        return "DocumentReplication {" +
+                "replicator=" + replicator +
+                ", is pushing =" + pushing +
+                ", document id =" + docId +
+                ", error code =" + error.getCode()+
+                ", error domain=" + error.getDomain() +
+                ", doc is deleted =" + isDeleted +
```
pasin: Should we include isAccessRemoved here as well?

shared/src/main/java/com/couchbase/lite/DocumentReplication.java
Sandychuang8: yes, isAccessRemoved value need to be retrieved which will be done in other ticket. I will also add test case(s) for this.

shared/src/main/java/com/couchbase/lite/DocumentReplication.java
Sandychuang8: I will add isAccessRemoved value in string when I am able to get isAccessRemoved in other ticket (PR).

shared/src/main/java/com/couchbase/lite/DocumentReplication.java
pasin: OK

shared/src/main/java/com/couchbase/lite/DocumentReplication.java
Sandychuang8: Looks like default errorCode value is 0.

android/CouchbaseLite/src/ce/java/com/couchbase/lite/ReplicatorConfiguration.java
```diff
@@ -43,6 +43,7 @@
     static final String kC4ReplicatorCheckpointInterval = "checkpointInterval"; // How often to checkpoint, in seconds; number
     static final String kC4ReplicatorOptionRemoteDBUniqueID = "remoteDBUniqueID"; // How often to checkpoint, in seconds; number
     static final String kC4ReplicatorResetCheckpoint = "reset"; // reset remote checkpoint
+    static final String kC4ReplicatorOptionProgressLevel = "progress";  //< If >=1, notify on every doc; if >=2, on every attachment (int)
```
pasin: I have just noticed this. Please remove as it is not used.

shared/src/main/java/com/couchbase/lite/AbstractReplicator.java
```diff
@@ -40,6 +42,7 @@
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
 
+import static com.couchbase.lite.ReplicatorConfiguration.kC4ReplicatorOptionProgressLevel;
```
Sandychuang8: It is used to set the progress level...

PR #1801, opened 2018-12-20
Implement new CBL logging logic

borrrden:


android/CouchbaseLite/src/androidTest/java/com/couchbase/lite/DatabaseTest.java
```diff
@@ -321,6 +321,7 @@ public void testGetExistingDocWithIDFromDifferentDBInstance() throws CouchbaseLi
     @Test
     public void testGetExistingDocWithIDInBatch() throws CouchbaseLiteException {
         final int NUM_DOCS = 10;
+        Database.getLog().getConsole().setLevel(LogLevel.VERBOSE);
```
pasin: Is this for testing if the console log works? Can we delete it if we don't need it?

android/CouchbaseLite/src/androidTest/java/com/couchbase/lite/LogTest.java
```diff
@@ -0,0 +1,168 @@
...
+        for(File log : logPath.listFiles()) {
+            BufferedReader fin = new BufferedReader(new FileReader(log));
+            int lineCount = 0;
+            while(fin.readLine() != null) {
+                lineCount++;
+            }
+
+            // Two meta lines per log, so the actual logging lines is X - 2
+            if(log.getAbsolutePath().contains("verbose")) {
+                assertEquals(3, lineCount);
```
pasin: This is no correct anymore. It should be 2, 3, 4, 5 now.

android/CouchbaseLite/src/main/java/com/couchbase/lite/ConsoleLogger.java
```diff
@@ -0,0 +1,96 @@
...
+
+    ConsoleLogger() {
+        _domains.add(LogDomain.ALL);
+    }
+
+    public ArrayList<LogDomain> getDomains() {
+        return _domains;
+    }
+
+    public void setDomains(ArrayList<LogDomain> domains) {
```
pasin: I think we should guard null domains here.

android/CouchbaseLite/src/main/java/com/couchbase/lite/ConsoleLogger.java
```diff
@@ -0,0 +1,96 @@
...
+        _domains = domains;
+    }
+
+    public void setLevel(LogLevel level) {
+        if(level == null) {
```
pasin: Maybe we should guard null LogLevel? or null would mean `NONE`?

How does .NET handle this?

shared/src/main/java/com/couchbase/lite/FileLogger.java
```diff
@@ -0,0 +1,159 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+public final class FileLogger implements Logger {
+    private LogLevel _level = LogLevel.INFO;
+    private String _directory;
+    private boolean _customDirectory;
+    private boolean _hasConfigChanges;
+    private int _maxRotateCount = 1;
+    private long _maxSize = 1024;
```
pasin: This should be 500 * 1024.

shared/src/main/java/com/couchbase/lite/FileLogger.java
```diff
@@ -0,0 +1,159 @@
...
+        _hasConfigChanges = true;
+    }
+
+
+    public String getDirectory() {
+        return _directory;
+    }
+
+    public void setDirectory(String directory) {
+        if(directory == null) {
```
pasin: Wondering if we should just guard null here?

shared/src/main/java/com/couchbase/lite/LogDomain.java
```diff
@@ -21,5 +21,5 @@
  * Log domain. The log domains here are tentative and subject to change.
  */
 public enum LogDomain {
-    ALL, DATABASE, QUERY, REPLICATOR, NETWORK
+    ALL, COUCHBASE, DATABASE, QUERY, REPLICATOR, NETWORK
```
pasin: You told me once but I forgot. When will the COUCHBASE domain be used?

After thought, can we use DATABASE domain for that so that we don't need to add a new domain here?

shared/src/main/java/com/couchbase/lite/internal/support/Log.java
```diff
@@ -385,49 +342,72 @@ public static void e(String tag, String formatString, Throwable tr, Object... ar
...
+        String msg;
+        try {
+            msg = String.format(Locale.ENGLISH, formatString, args);
+        } catch (Exception e) {
+            msg = String.format(Locale.ENGLISH, "Unable to format log: %s (%s)", formatString, e.toString());
         }
+        sendToLoggers(LogLevel.ERROR, tag, msg);
     }
 
     public static void setLogLevel(LogDomain domain, LogLevel level) {
```
pasin: We should make it deprecated.

shared/src/main/java/com/couchbase/lite/internal/support/Log.java
```diff
@@ -385,49 +342,72 @@ public static void e(String tag, String formatString, Throwable tr, Object... ar
...
             case NETWORK:
-                enableLogging(C4LogDomain.BLIP, level.getValue());
-                enableLogging(WEB_SOCKET, level.getValue());
+                enableLogging(C4LogDomain.BLIP, actualLevel);
+                enableLogging(WEB_SOCKET, actualLevel);
                 break;
         }
     }
 
     public static void enableLogging(String tag, int logLevel) {
```
pasin: Do you know when this method is used? It's public as well.

shared/src/main/java/com/couchbase/lite/internal/support/Log.java
```diff
@@ -385,49 +342,72 @@ public static void e(String tag, String formatString, Throwable tr, Object... ar
...
+
+    private static void sendToLoggers(LogLevel level, String tag, String msg) {
+        boolean fileSucceeded = false;
+        boolean consoleSucceeded = false;
+        try {
+            LogDomain domain = LogDomain.valueOf(tag);
+            Database.getLog().getFile().log(level, domain, msg);
+            fileSucceeded = true;
+            Database.getLog().getConsole().log(level, domain, msg);
+            consoleSucceeded = false;
```
pasin: should it set to true?

shared/src/main/java/com/couchbase/lite/internal/support/Log.java
```diff
@@ -385,49 +342,72 @@ public static void e(String tag, String formatString, Throwable tr, Object... ar
...
+            Logger custom = Database.getLog().getCustom();
+            if(custom != null) {
+                custom.log(level, domain, msg);
+            }
+        } catch(Exception e) {
+            if(fileSucceeded) {
```
pasin: I'm not sure if I understand this here. Should it be if (!fileSucceeded)?

shared/src/main/java/com/couchbase/lite/internal/support/Log.java
```diff
@@ -385,49 +342,72 @@ public static void e(String tag, String formatString, Throwable tr, Object... ar
...
+                Database.getLog().getFile().log(LogLevel.ERROR, LogDomain.DATABASE, e.toString());
+            }
+
+            if(consoleSucceeded) {
```
pasin: I'm not sure if I understand this here. Should it be if (!consoleSucceeded)?

shared/src/main/java/com/couchbase/lite/LogDomain.java
pasin: I couldn't find where it get used in this PR as well.

shared/src/main/java/com/couchbase/lite/internal/support/Log.java
borrrden: No, if fileSucceeded is false it means something is wrong with the file logger, and logging to it would cause another exception inside of the catch.  Basically, I want to log the error to all the loggers that are still working.

shared/src/main/java/com/couchbase/lite/LogDomain.java
borrrden: COUCHBASE will be the default domain for unmapped LiteCore domains.  It is used in the JNI (coincidentally one of the problems Sandy ran into was "COUCHBASE not defined")

pasin: Can we do DATABASE instead so we don't need to add a new domain?

android/CouchbaseLite/src/main/java/com/couchbase/lite/ConsoleLogger.java
pasin: So let to NONE if null is set?

borrrden: It's up to you, this is a situation specific to the Java language.

shared/src/main/java/com/couchbase/lite/FileLogger.java
pasin: Oh I see here, I think we don't allow to set null, users can't set it back to default. I think I need to change iOS implementation for this.

android/CouchbaseLite/src/main/java/com/couchbase/lite/ConsoleLogger.java
pasin: Then let's do NONE.

shared/src/main/java/com/couchbase/lite/internal/support/Log.java
borrrden: I don't think I added this.  It's used for LiteCore log levels.

shared/src/main/java/com/couchbase/lite/LogDomain.java
pasin: This might now answer the question. The weird part is that as a user I don't know when I should include COUCHBASE in the domain list. It seems like it is a domain that should be included all the time because it's something that might get logged randomly?

pasin: Is it something that can be handled case by case to manually catch and map to one of the existing domains?

borrrden: I guess it's more of a side effect of giving the user the choice at all.  How many will simply just change the log level and not bother with domains?  Ultimately I think they should all be removed (maybe even internally) to make things simpler so I guess I can put up with Database being "odd" for now and we can answer the question properly in 3.0.

pasin: Totally agree.

shared/src/main/java/com/couchbase/lite/internal/support/Log.java
pasin: OK.

shared/src/main/java/com/couchbase/lite/LogDomain.java
borrrden: I still want to come up with a mapping so that not everything gets dumped into database.  I'll write one today.

Issue #1084, opened 2017-03-03
2.0: Define 2.0 Java API interface from the spec


hideki:
https://github.com/couchbaselabs/couchbase-lite-apiv2

PR #1806, opened 2018-12-27
#1084 Use lazy deserialization for replication filters.

Sandychuang8:


shared/src/main/java/com/couchbase/lite/MutableDocument.java
```diff
@@ -51,7 +52,7 @@ public MutableDocument() {
      * @param id the document ID.
      */
     public MutableDocument(String id) {
-        super(null, id != null ? id : createUUID(), null);
+        super(null, id != null ? id : createUUID(), (C4Document)null);
```
jayahariv: In Java, do we have something like this? 
```suggestion
        super(null, id ?: createUUID(), (C4Document)null); 
        // OR 
        super(null, id ?? createUUID(), (C4Document)null); 
```

shared/src/main/java/com/couchbase/lite/AbstractReplicator.java
```diff
@@ -677,8 +677,7 @@ private void notifyDocumentEnded(DocumentReplication update)
     }
 
     private boolean validationFunction(String docID, int flags, long dict, boolean isPush) {
-        Map<String, Object> dictionary = new FLDict(dict).asDict();
-        MutableDocument document = new MutableDocument(docID, dictionary);
+        Document document = new Document(config.getDatabase(), docID, new FLDict(dict));
```
jayahariv: is there any impact of changing the type from MutableDocument to Document?

Sandychuang8: MutableDocument is derived from Document and since it now requires to use lazy deserialization for replication filters so initialize from Document is probably the right way..

Issue #1810, opened 2019-01-04
Update Replication Filter and Document Replication Eventing API


pasin:
Update the API according to the spec [here](https://github.com/couchbaselabs/couchbase-lite-apiv2/pull/237).
* [x] DocumentFlags
* [x] ReplicationFilter
* [x] ReplicatedDocument

PR #1812, opened 2019-01-04
Update Replication Filter and Document Replication Eventing API.

Sandychuang8:
Update Replication Filter and Document Replication Eventing API. Add new DocumentFlags class takes in isDeleted flag and isAccessRemoved flag. ReplicatedDocument class instead of taking isDeleted flag and isAccessRemoved flags, now it takes DocumentFlags object contains the two flags.


shared/src/main/java/com/couchbase/lite/DocumentFlags.java
```diff
@@ -0,0 +1,47 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+package com.couchbase.lite;
+
+import com.couchbase.litecore.C4Constants;
+
+public class DocumentFlags {
```
pasin: Can this be implemented as a kind of bitwise enum? I remembered that Java has EnumSet but I'm not sure if it's appropriate for this or not.

@borrrden I think you have looked into Java Bitwise enum before. What is your opinion on Java implementation?

borrrden: It looks like the standard practice is to use `EnumSet` to accomplish this -> https://stackoverflow.com/questions/6073347/best-practice-of-using-flags-in-java-method

shared/src/main/java/com/couchbase/lite/DocumentFlag.java
```diff
@@ -0,0 +1,16 @@
+package com.couchbase.lite;
```
pasin: Please add file header and license?

shared/src/main/java/com/couchbase/lite/DocumentFlag.java
```diff
@@ -0,0 +1,16 @@
...
+
+public enum DocumentFlag {
```
pasin: Please add Java Doc.

shared/src/main/java/com/couchbase/lite/ReplicatedDocument.java
```diff
@@ -58,17 +68,10 @@ public String getID() {
...
-        return isDeleted;
-    }
-
-    /**
-     * The current access removed status of the document.
+     * The current status flag of the document. eg. deleted, access removed
      */
-    public boolean isAccessRemoved()  {
-        return isAccessRemoved;
+    public Set<DocumentFlag> flags() {
```
pasin: Should it be EnumSet instead of Set (ReplicationFilter interface uses EnumSet)? 

Could you do some research that normally it should be Set or EnumSet?

Sandychuang8: Most EnumSet example usages is EnumSet.

Sandychuang8:
close #1810

Issue #1814, opened 2019-01-09
Change domains type of the Console Logger to EnumSet


pasin:
We should change domains type of the Console Logger to EnumSet to be compatible with the other platforms that use bitwise enum.

PR #1816, opened 2019-01-10
Change domains type of the Console Logger to EnumSet.

Sandychuang8:
#1814 Change domains type of the Console Logger to EnumSet to match other platforms that use bitwise enum.

### DIFF
--- a/android/CouchbaseLite/src/main/java/com/couchbase/lite/ConsoleLogger.java
+++ b/android/CouchbaseLite/src/main/java/com/couchbase/lite/ConsoleLogger.java
@@ -21,7 +21,6 @@ import android.util.Log;
 
 import com.couchbase.litecore.C4Log;
 
-import java.util.ArrayList;
 import java.util.EnumSet;
 
 /**
@@ -31,14 +30,12 @@ import java.util.EnumSet;
  */
 public final class ConsoleLogger implements Logger {
     private LogLevel _level = LogLevel.WARNING;
-    private EnumSet<LogDomain> _domains = EnumSet.noneOf(LogDomain.class);
+    private EnumSet<LogDomain> _domains = EnumSet.allOf(LogDomain.class);
 
     //---------------------------------------------
     // Constructor should not be exposed (singleton)
     //---------------------------------------------
-    ConsoleLogger() {
-        _domains.addAll(EnumSet.allOf(LogDomain.class));
-    }
+    ConsoleLogger() { }
 
     /**
      * Gets the domains that will be considered for writing to
@@ -102,8 +99,9 @@ public final class ConsoleLogger implements Logger {
 
     @Override
     public void log(LogLevel level, LogDomain domain, String message) {
-        if(level.compareTo(_level) < 0 || (!_domains.contains(domain)
-                && !_domains.containsAll(EnumSet.allOf(LogDomain.class)))) {
+        if(level.compareTo(_level) < 0
+                || (!_domains.contains(domain)
+                && !_domains.contains(LogDomain.ALL))) {
             return;
         }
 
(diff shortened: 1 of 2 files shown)


android/CouchbaseLite/src/main/java/com/couchbase/lite/ConsoleLogger.java
```diff
@@ -30,13 +31,13 @@
...
     private LogLevel _level = LogLevel.WARNING;
-    private ArrayList<LogDomain> _domains = new ArrayList<>();
+    private EnumSet<LogDomain> _domains = EnumSet.noneOf(LogDomain.class);
 
     //---------------------------------------------
     // Constructor should not be exposed (singleton)
     //---------------------------------------------
     ConsoleLogger() {
-        _domains.add(LogDomain.ALL);
+        _domains.addAll(EnumSet.allOf(LogDomain.class));
```
pasin: As we have ALL, I think you can just add only ALL here.

android/CouchbaseLite/src/main/java/com/couchbase/lite/ConsoleLogger.java
```diff
@@ -101,7 +102,8 @@ public LogLevel getLevel() {
 
     @Override
     public void log(LogLevel level, LogDomain domain, String message) {
-        if(level.compareTo(_level) < 0 || (!_domains.contains(domain) && !_domains.contains(LogDomain.ALL))) {
+        if(level.compareTo(_level) < 0 || (!_domains.contains(domain)
+                && !_domains.containsAll(EnumSet.allOf(LogDomain.class)))) {
```
pasin: I think the original logic is already correct.

android/CouchbaseLite/src/main/java/com/couchbase/lite/ConsoleLogger.java
```diff
@@ -30,13 +31,13 @@
  */
 public final class ConsoleLogger implements Logger {
     private LogLevel _level = LogLevel.WARNING;
-    private ArrayList<LogDomain> _domains = new ArrayList<>();
+    private EnumSet<LogDomain> _domains = EnumSet.noneOf(LogDomain.class);
```
pasin: I think you could initialize the _domains here to LogDomain.ALL then we don't need to do anything in the ConsoleLogger() constructor.

android/CouchbaseLite/src/main/java/com/couchbase/lite/ConsoleLogger.java
```diff
@@ -22,6 +22,7 @@
 import com.couchbase.litecore.C4Log;
 
 import java.util.ArrayList;
```
pasin: Do we still need ArrayList?

android/CouchbaseLite/src/main/java/com/couchbase/lite/ConsoleLogger.java
```diff
@@ -30,22 +30,20 @@
  */
 public final class ConsoleLogger implements Logger {
     private LogLevel _level = LogLevel.WARNING;
-    private ArrayList<LogDomain> _domains = new ArrayList<>();
+    private EnumSet<LogDomain> _domains = EnumSet.allOf(LogDomain.class);
```
pasin: I think we should do `private EnumSet<LogDomain> _domains = EnumSet.of(LogDomain.ALL);`

shared/src/main/java/com/couchbase/lite/LogDomain.java
```diff
@@ -17,9 +17,51 @@
 //
 package com.couchbase.lite;
 
+import java.util.EnumSet;
+
 /**
- * Log domain. The log domains here are tentative and subject to change.
+ * Log domain
  */
 public enum LogDomain {
```
pasin: Please revert LogDomain Enum to the original as there is to do bitwise at the Enum level. EnumSet has already done the job.

```Java
public enum LogDomain {
    ALL, DATABASE, QUERY, REPLICATOR, NETWORK
}
```

PR #1823, opened 2019-01-18
Change logging API to match spec

### LINKED ISSUE #1821: Remove log rotation change batching
The batching of things like the file logging max size, max rotate count, etc is prone to races and generally over complicated.  This is dependent on https://github.com/couchbase/couchbase-lite-core/issues/657.

borrrden:
Closes #1821

### DIFF
--- a/shared/src/main/java/com/couchbase/lite/FileLogger.java
+++ b/shared/src/main/java/com/couchbase/lite/FileLogger.java
@@ -32,14 +32,8 @@ import java.util.Map;
  */
 public final class FileLogger implements Logger {
     private LogLevel _level = LogLevel.INFO;
-    private String _directory;
-    private boolean _customDirectory;
-    private boolean _hasConfigChanges;
-    private int _maxRotateCount = 1;
-    private long _maxSize = 500 * 1024;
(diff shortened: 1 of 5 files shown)


shared/src/main/java/com/couchbase/lite/FileLogger.java
```diff
@@ -167,7 +71,11 @@ public void setDirectory(String directory) {
      *
      * @param level The maximum level to include in the logs
      */
-    public void setLevel(LogLevel level) {
+    public void setLevel(LogLevel level) throws IllegalStateException {
```
pasin: For consistency, we have never declared to throw IllegalStateException for null checking. I think we should remove it for consistency.

borrrden: Interesting.  I thought it was required to add this.....there is much I don't know about Java.  I'll try removing this stuff and see if it complains.

pasin: It's not required to declare runtime exception as it's not required to catch it. I will remove the declaration and push to this PR so that we can just merge.

PR #1824, opened 2019-01-19
Change to support PredictiveQuery implementation

pasin:
* Made FunctionAbstract and IndexBuilderAbstract so that we can have predictive query related feature only available in EE.

* Added Expression.map(), Expression.list(), Parameters.dictionary(), Parameters.array(), and Parameters.blob().

* Updated MValueDelegate to support converting byte[] to Blob

* Fixed JSONUtil to support null.

* Fixed mis-spelled FunctionExpresson to FunctionExpression.

* Moved some util methods from QueryTest to BaseTest so PredictiveQueryTest in EE can use those methods.


shared/src/main/java/com/couchbase/lite/MValueDelegate.java
```diff
@@ -44,6 +44,8 @@ public Object toNative(MValue mv, MCollection parent, AtomicBoolean cacheIt) {
             case kFLDict:
                 cacheIt.set(true);
                 return mValueToDictionary(mv, parent);
+            case kFLData:
+                return new Blob("application/octet-stream", value.asData());
```
borrrden: Need to talk about this later since this case will no longer be hit after the blob metadata fix.

pasin:
@borrrden any more comments beside the above?

borrrden:
No, nothing else.

shared/src/main/java/com/couchbase/lite/MValueDelegate.java
pasin: This will be fixed in a separate PR.

PR #1827, opened 2019-01-23
Fixes #1825 NULL parameters should be validated and throw Runtime Exc…

### LINKED ISSUE #1825: Validate null parameters in Query's components
NULL parameters should be validated and throw Runtime Exception if non-null is required.

Example: Having(null) should throw Exception.

Sandychuang8:
Fixes #1825 NULL parameters should be validated and throw Runtime Exception.


pasin:
Not only GroupBy but we should review other API and make sure to validate them as well. I noticed that non of the Query API have validated NULL.

Sandychuang8:
There are more than this one case need to validate null. So close this PR and will create a new PRs.

Issue #1808, opened 2019-01-04
Iridium Logging Test Cases


borrrden:
See the CBL Logging test plan for details on each item

- [ ] Log availability at default location
- [x] Logging format with default behavior
- [x] Enabling Text logging
- [x] Logs for each log level
- [x] Max size log rotation
- [x] Disable logging
- [x] Re-enable logging
- [x] log file name
- [x] System configuration Snippet at the start of log
- [x] Custom logging
- [x] Custom logging log level
- [x] Disable custom logging
- [x] re-enable custom logging

jayahariv:
File logging, will add unit tests after the new change is included

jayahariv:
Log availability at default location: no checking done, since we are getting the directory from the users.

jayahariv:
Disable custom logging can be checked inside the Re-enable custom logging. As the re-enable custom logging include a disabling and validation for not printing the logs. So will go ahead and mark disable as done.

PR #1829, opened 2019-01-23
test: add remaining file logging unit tests

jayahariv:
* check for file max size
* check for disable, reenable logging
* check for file header.

* note: file max size is not working, as the debug file is not printing anything in it.
* note: file header also not printing, seems like user agent is not passed to FileOptions in JNI

link: #1808

### DIFF
--- a/android/CouchbaseLite/src/androidTest/java/com/couchbase/lite/LogTest.java
+++ b/android/CouchbaseLite/src/androidTest/java/com/couchbase/lite/LogTest.java
@@ -1,6 +1,7 @@
 package com.couchbase.lite;
 
 import android.content.Context;
+import android.content.pm.ApplicationInfo;
 import android.support.test.InstrumentationRegistry;
 
 import com.couchbase.lite.internal.support.Log;
@@ -138,11 +139,16 @@ public class LogTest {
 
                 int maxRotateCount = config.getMaxRotateCount();
                 int totalFilesShouldBeInDirectory = (maxRotateCount + 1) * 5;
+                boolean isDebug = (
+                        (context.getApplicationInfo().flags & ApplicationInfo.FLAG_DEBUGGABLE) != 0
+                );
+                if (!isDebug) {
+                    totalFilesShouldBeInDirectory -= 1;
+                }
                 int totalLogFilesSaved = path.listFiles().length;
                 assertEquals(totalFilesShouldBeInDirectory, totalLogFilesSaved);
             }
         });
-        emptyDirectory(path.getAbsolutePath());
     }
 
     @Test
@@ -174,7 +180,6 @@ public class LogTest {
                 }
             }
         });
-        emptyDirectory(path.getAbsolutePath());
     }
 
     @Test
@@ -226,10 +231,9 @@ public class LogTest {
                 }
             }
         });
-        emptyDirectory(path.getAbsolutePath());
     }
 
-    @Test @Ignore
+    @Test
     public void testFileLoggingHeader() {
         final File path = new File(
                 context.getCacheDir().getAbsolutePath(),
@@ -255,7 +259,6 @@ public class LogTest {
                 }
             }
         });
-        emptyDirectory(path.getAbsolutePath());
     }
 
     private void testWithConfiguration(LogLevel level, LogFileConfiguration config, Runnable r) {
@@ -276,7 +279,6 @@ public class LogTest {
             for (File log : f.listFiles()) {
                 log.delete();
             }
-            f.delete();
         }
         return path;
     }
--- a/android/CouchbaseLite/src/androidTest/java/com/couchbase/lite/LogTest.java
+++ b/android/CouchbaseLite/src/androidTest/java/com/couchbase/lite/LogTest.java
@@ -6,15 +6,22 @@ import android.support.test.InstrumentationRegistry;
 import com.couchbase.lite.internal.support.Log;
 
 import org.junit.Before;
+import org.junit.Ignore;
 import org.junit.Test;
 
 import java.io.BufferedReader;
 import java.io.File;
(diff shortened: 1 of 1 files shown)


android/CouchbaseLite/src/androidTest/java/com/couchbase/lite/LogTest.java
```diff
@@ -111,6 +118,146 @@ public void testEnableAndDisableCustomLogging() {
         assertEquals(4, customLogger.getLines().size());
     }
 
+    @Test @Ignore
```
jayahariv: ignored since the debug logs are not printing anything in it

android/CouchbaseLite/src/androidTest/java/com/couchbase/lite/LogTest.java
```diff
@@ -111,6 +118,146 @@ public void testEnableAndDisableCustomLogging() {
...
+
+                } catch(Exception e) {
+                    fail("Exception during test callback " + e.toString());
+                }
+            }
+        });
+        emptyDirectory(path.getAbsolutePath());
+    }
+
+    @Test @Ignore
```
jayahariv: ignored since the [updateConfig](https://github.com/couchbase/couchbase-lite-android/blob/1b96876a3fb00a3dcbc634c1cd9778a2a5f6dd85/shared/src/main/java/com/couchbase/lite/FileLogger.java#L100) method is not passing in the user agent info. 

[native_c4.cc](https://github.com/couchbase/couchbase-lite-core/blob/31266c807a208ffa1569a339fd1954dc0bbb7cb0/Java/jni/native_c4.cc#L148)

android/CouchbaseLite/src/androidTest/java/com/couchbase/lite/LogTest.java
```diff
@@ -129,10 +276,24 @@ private static String emptyDirectory(String path){
             for (File log : f.listFiles()) {
                 log.delete();
             }
+            f.delete();
```
jayahariv: folder was not cleaned up.

android/CouchbaseLite/src/androidTest/java/com/couchbase/lite/LogTest.java
borrrden: So why ignore instead of adjusting the test?  For release mode it's true no debug logs should be written but in debug mode the debug logs should be written.

jayahariv: waiting to discuss with Pasin tomorrow, and need to understand how the debug and release works in Android.

pasin: * Android Studio has a separate option/button for run (release build type) and debug. To enable debug build type, select debug.
* I understand that when running test from Jenkins using gradle command, the release build type will be used.
* As @borrrden suggested, I think we should modify the test to ignore or not checking debug.

android/CouchbaseLite/src/androidTest/java/com/couchbase/lite/LogTest.java
jayahariv: then we can keep it the folder, to re-use.

android/CouchbaseLite/src/androidTest/java/com/couchbase/lite/LogTest.java
jayahariv: I used below code to check whether the test is in debug or not. and I am getting `isDebug=true` but still, `cbl_debug_xxx` file only with heading. 

any suggestions

```
boolean isDebug = (
                        (context.getApplicationInfo().flags & ApplicationInfo.FLAG_DEBUGGABLE) != 0
```

android/CouchbaseLite/src/androidTest/java/com/couchbase/lite/LogTest.java
```diff
@@ -130,9 +280,22 @@ private static String emptyDirectory(String path){
                 log.delete();
             }
         }
-
         return path;
     }
+
+    private static void writeOneKiloByteOfLog() {
+        String message = "11223344556677889900"; // ~43 bytes
+        for (int i = 0; i < 24; i++) { // 24 * 43 = 1032
```
jayahariv:
```suggestion
        for (int i = 0; i < 24; i++) { // 24 * 43 = ~1032 bytes
```

android/CouchbaseLite/src/androidTest/java/com/couchbase/lite/LogTest.java
pasin: I guessed CMAKE_BUILD_TYPE doesn't get set correctly according to the android build type.

We may need to explicitly set the CMAKE_BUILD_TYPE to the right value in https://github.com/couchbase/couchbase-lite-android/blob/master/android/CouchbaseLite/build.gradle#L86-L97.

jayahariv: i will merge the PR, and create a separate issue to fix the test `testFileLoggingMaxSize`. So that we don't need to keep other tests.

jayahariv: referencing issue #1835 and resolving this conversation.

jayahariv:
test failed on different case. 

```
testSetBoolean[test(AVD) - 4.4.2] FAILED 
Test failed to run to completion. Reason: 'Instrumentation run failed due to 'Native crash''. Check device logcat for details
Tests on test(AVD) - 4.4.2 failed: Instrumentation run failed due to 'Native crash'
```

PR #1841, opened 2019-01-25
Add null check with NonNull annotations in Query Index classes.

Sandychuang8:
Add null check with NonNull annotations in Query Index classes.

### DIFF
--- a/shared/src/main/java/com/couchbase/lite/FullTextIndexItem.java
+++ b/shared/src/main/java/com/couchbase/lite/FullTextIndexItem.java
@@ -17,6 +17,8 @@
 //
 package com.couchbase.lite;
 
+import android.support.annotation.NonNull;
+
 /**
  * Full-text Index Item.
  */
@@ -32,8 +34,13 @@ public final class FullTextIndexItem {
      *
      * @param property A property used to perform the match operation against with.
      * @return The full-text search index item.
+     * @throws IllegalArgumentException when property is null.
      */
-    public static FullTextIndexItem property(String property) {
+    public static FullTextIndexItem property(@NonNull String property) {
+
+        if(property == null) {
+            throw new IllegalArgumentException("property is null");
+        }
         return new FullTextIndexItem(Expression.property(property));
     }
 }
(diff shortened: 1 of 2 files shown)


shared/src/main/java/com/couchbase/lite/FullTextIndexItem.java
```diff
@@ -32,8 +34,13 @@ private FullTextIndexItem(Expression expression) {
      *
      * @param property A property used to perform the match operation against with.
      * @return The full-text search index item.
+     * @throws IllegalArgumentException when property is null.
      */
-    public static FullTextIndexItem property(String property) {
+    public static FullTextIndexItem property(@NonNull String property) {
+
+        if(property == null) {
+            throw new IllegalArgumentException("property is null");
```
pasin: Summary:

1. Removed `@throws IllegalArgumentException when alias is null.` as it is not required to declare throwing a runtime exception. We haven't declared this in other places e.g. Database class as well.

2. Remove an extra empty line before `if(<variable> == null) {`.

3. Add a space between if and (. Should be `if (<variable> == null)`.

4. Add . (full stop) at the end of "... is null" -> `... is null.`

5. Do the same for the other methods.

shared/src/main/java/com/couchbase/lite/ValueIndexItem.java
```diff
@@ -32,8 +34,13 @@ private ValueIndexItem(Expression expression) {
...
      * @param property the property name
      * @return The value index item
+     * @throws IllegalArgumentException when property is null.
      */
-    public static ValueIndexItem property(String property) {
+    public static ValueIndexItem property(@NonNull String property) {
+
+        if(property == null) {
+            throw new IllegalArgumentException("property is null");
+        }
```
pasin: Summary:

1. Removed `@throws IllegalArgumentException when alias is null.` as it is not required to declare throwing a runtime exception. We haven't declared this in other places e.g. Database class as well.

2. Remove an extra empty line before `if(<variable> == null) {`.

3. Add a space between if and (. Should be `if (<variable> == null)`.

4. Add . (full stop) at the end of "... is null" -> `... is null.`

5. Do the same for the other methods.

Issue #1457, opened 2017-12-01
2.0: Improve code coverage rate to 80%.


hideki:
With Build 168, Coverage is 52%. 
http://mobile.jenkins.couchbase.com/view/Core_Based_Builds/job/couchbase-lite-android/168/artifact/reports/reports-android-24/coverage/debug/index.html

hideki:
## Notes

#### Unavoidable miss-coverage
- private constructor
```java
public class LiteCoreBridge {
    private LiteCoreBridge(){}
}
```

hideki:
60% - 12/13/2017
http://mobile.jenkins.couchbase.com/job/couchbase-lite-android/179/artifact/reports/reports-android-24/coverage/debug/index.html

jayahariv:
Reached 79% for EE 
Reached 76% for CE 

## EE
<img width="1169" alt="ee" src="https://user-images.githubusercontent.com/10448770/52880698-7e086900-3117-11e9-8f86-8d089faea308.png">

## CE
<img width="1139" alt="ce" src="https://user-images.githubusercontent.com/10448770/52880836-f111df80-3117-11e9-8819-38c49cc9c938.png">

Hence closing this for now.

PR #1860, opened 2019-02-06
test: add offline replicator tests

jayahariv:
* add replicator tests offline which will ran in CE edition.

link: #1457

### DIFF
--- a/android/CouchbaseLite/src/androidTest/java/com/couchbase/lite/ReplicatorOfflineTest.java
+++ b/android/CouchbaseLite/src/androidTest/java/com/couchbase/lite/ReplicatorOfflineTest.java
@@ -34,14 +34,10 @@ public class ReplicatorOfflineTest extends BaseReplicatorTest {
     }
 
     @Test
-    public void testOfflineStart() throws Exception {
-        Endpoint endpoint = getRemoteTargetEndpoint();
-        Replicator repl = new Replicator(makeConfig(
-                true,
-                false,
-                false,
-                endpoint
-        ));
+    public void testStopReplicatorAfterOffline() throws URISyntaxException, InterruptedException {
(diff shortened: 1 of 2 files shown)


android/CouchbaseLite/src/androidTest/java/com/couchbase/lite/ReplicatorOfflineTest.java
```diff
@@ -0,0 +1,87 @@
...
+        );
+        config.setContinuous(false);
+        repl = new Replicator(config);
+
+        thrown.expect(IllegalStateException.class);
+        repl.getConfig().setContinuous(true);
+    }
+
+    @Test
+    public void testOfflineStart() throws Exception {
```
jayahariv: This test is same as SyncGatewayTest: [testStopReplicatorAfterOffline](https://github.com/couchbase/couchbase-lite-android/blob/34d44d842d378e440519ea73e1d92ee4de63aaa2/android/CouchbaseLite/src/androidTest/java/com/couchbase/lite/ReplicatorWithSyncGatewayTest.java#L68)

Should we copy the same test to `ReplicatorOfflineTest` ?

pasin: Just leave it at the SyncGatewayTest.

pasin: I changed my mind. You can copy it here so that we could cover stop()?

jayahariv: so we don't want to duplicate same test in two places. even without sync gateway that test can be executed.

pasin: OK, I think you can move that test here.

PR #1863, opened 2019-02-07
test: add blob from file check

jayahariv:
* check blob from file
* check blob read(), read(toDifferentByteArray), skip() methods

link: #1457

### DIFF
--- a/android/CouchbaseLite/src/androidTest/java/com/couchbase/lite/BlobTest.java
+++ b/android/CouchbaseLite/src/androidTest/java/com/couchbase/lite/BlobTest.java
@@ -19,10 +19,14 @@ package com.couchbase.lite;
 
 import com.couchbase.lite.utils.IOUtils;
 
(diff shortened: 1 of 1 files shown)


android/CouchbaseLite/src/androidTest/java/com/couchbase/lite/BlobTest.java
```diff
@@ -34,6 +38,9 @@
     final static String kBlobTestBlob1 = "i'm blob";
     final static String kBlobTestBlob2 = "i'm blob2";
 
+    @Rule
+    public ExpectedException thrown = ExpectedException.none();
```
pasin: I'm wondering if we should move this to BaseTest.

jayahariv: that will avoid declaring in many places. we can do that in a separate PR.

PR #1887, opened 2019-03-20
Add Java PoC and fixes to make it work

bmeike:
Added a minimal PoC subproject in the folder `java`.  It is a pure java build of the code in the top level `shared` directory, along with minimal local code necessary to get it to work.

This is not a working project by any means.  It simply proves that all references to Android are gone from the shared code.

There were a few minor changes both in shared and platform-specific code, to get this to work.


android/CouchbaseLite/src/ce/java/com/couchbase/lite/ReplicatorConfiguration.java
```diff
@@ -1,7 +1,10 @@
+package com.couchbase.lite;
+
+import android.support.annotation.NonNull;
+
+
```
pasin: package and imports are after the license.

shared/src/main/java/com/couchbase/lite/AbstractReplicatorConfiguration.java
```diff
@@ -0,0 +1,405 @@
...
+
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+
+import com.couchbase.lite.internal.core.CBLVersion;
+
+
+/**
+ * Replicator configuration.
+ */
+public abstract class AbstractReplicatorConfiguration {
```
pasin: This class shouldn't be public. I understand that Java allows to override non-public abstract class to hide the class from public visibility.

android/CouchbaseLite/src/ce/java/com/couchbase/lite/ReplicatorConfiguration.java
```diff
@@ -15,394 +18,18 @@
...
-        this.continuous = config.continuous;
-        this.authenticator = config.authenticator;
-        this.pinnedServerCertificate = config.pinnedServerCertificate;
-        this.headers = config.headers;
-        this.channels = config.channels;
-        this.documentIDs = config.documentIDs;
-        this.pullFilter = config.pullFilter;
-        this.pushFilter = config.pushFilter;
+public final class ReplicatorConfiguration extends AbstractReplicatorConfiguration {
+    public ReplicatorConfiguration(@NonNull AbstractReplicatorConfiguration config) {
```
pasin: The signature should be `public ReplicatorConfiguration(@NonNull ReplicatorConfiguration config)` instead.

java/src/main/java/com/couchbase/lite/CouchbaseLite.java
```diff
@@ -0,0 +1,37 @@
...
+//
+package com.couchbase.lite;
+
+import android.support.annotation.NonNull;
+
+
+public final class CouchbaseLite {
+    private CouchbaseLite() {}
+
+    public static void init() { }
```
pasin: I understand that the init() needs `directory` as a parameter.

java/src/main/java/com/couchbase/lite/FileLogger.java
```diff
@@ -0,0 +1,156 @@
...
+import com.couchbase.lite.internal.support.Log;
+
+
+/**
+ * A logger for writing to a file in the application's storage so
+ * that log messages can persist durably after the application has
+ * stopped or encountered a problem.  Each log level is written to
+ * a separate file.
+ */
+public final class FileLogger implements Logger {
```
pasin: It seems like we can put FileLogger in the shared code.

shared/src/main/java/com/couchbase/lite/AbstractReplicatorConfiguration.java
```diff
@@ -0,0 +1,405 @@
+//
+// ReplicatorConfiguration.java
```
pasin: AbstractReplicatorConfiguration

java/src/main/java/com/couchbase/lite/ReplicatorConfiguration.java
```diff
@@ -0,0 +1,39 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+package com.couchbase.lite;
+
+import android.support.annotation.NonNull;
+
+
+public final class ReplicatorConfiguration extends AbstractReplicatorConfiguration {
```
pasin: Look exactly like the Android's version. Not sure why we need one for Java and one for Android.

java/src/main/java/com/couchbase/lite/internal/replicator/CBLWebSocket.java
```diff
@@ -0,0 +1,32 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+package com.couchbase.lite.internal.replicator;
+
+import java.net.URISyntaxException;
+import java.security.GeneralSecurityException;
+import java.util.Map;
+
+public class CBLWebSocket extends AbstractCBLWebSocket {
```
pasin: Do we need a separate version of CBLWebSocket between Java and Android? Does OKHTTP also support Java?

shared/src/main/java/com/couchbase/lite/AbstractReplicatorConfiguration.java
```diff
@@ -0,0 +1,405 @@
...
+    /**
+     * Sets the authenticator to authenticate with a remote target server.
+     * Currently there are two types of the authenticators,
+     * BasicAuthenticator and SessionAuthenticator, supported.
+     *
+     * @param authenticator The authenticator.
+     * @return The self object.
+     */
+    @NonNull
+    public ReplicatorConfiguration setAuthenticator(@NonNull Authenticator authenticator) {
```
pasin: This is interesting and I don't know how it works. ReplicatorConfiguration extends from AbstractReplicatorConfiguration but the methods in ReplicatorConfiguration returns an object of type ReplicatorConfiguration. This sounds like circular dependency.

Should this method return `AbstractReplicatorConfiguration` instead? BUT this means that ReplicatorConfiguration class will need to override each single method to return ReplicatorConfiguration.

shared/src/main/java/com/couchbase/lite/AbstractReplicatorConfiguration.java
bmeike: It has to be public.  All those constants are public.  It would be insanity to define the multiple times (as used to be the case).  What is the problem with making it public.  All of its contents used to be in a public class?

android/CouchbaseLite/src/ce/java/com/couchbase/lite/ReplicatorConfiguration.java
bmeike: Why?  This change does not affect the API?

java/src/main/java/com/couchbase/lite/CouchbaseLite.java
bmeike: Maybe.  If we do a java product we can worry about that.  This is a PoC.

java/src/main/java/com/couchbase/lite/FileLogger.java
bmeike: That seems dangerous to me.  I think there is every reason to believe that the Java file logger will, eventually, diverge, substantially, from the code that is currently there.  Please remember, this is the simplest possible proof of concept.

shared/src/main/java/com/couchbase/lite/AbstractReplicatorConfiguration.java
pasin: I think that would work.

java/src/main/java/com/couchbase/lite/FileLogger.java
pasin: The reason that I called this out because pretty all of the file logger logic has been in the Lite Core. The only thing that is required from the platform the configuration.

pasin: But as this is a POC, we could leave it out and see if we actually need to separate it out later.

bmeike: Ok... Yeah, I *hope* you are right and that we can keep the file logic in the `CouchbaseLite` class.  I *wasn't* able to do that for the `ConsoleLogger`, though, so I'm just trying to be safe.

Issue #1769, opened 2018-10-23
CBL 2.x is not compatible with x86_64 devices


MatFl:
Please provide the native libraries for x86 64bit CPUs, many devices with e.g. Intel Atoms require this!

bmeike:
Duplicated in https://github.com/couchbase/couchbase-lite-android/issues/1795

bmeike:
Duplicated in https://github.com/couchbase/couchbase-lite-android/issues/1878

PR #1892, opened 2019-03-21
Fix Issue #1769: produce version for x86_64

bmeike:


android/CouchbaseLite/build.gradle
```diff
@@ -67,6 +67,11 @@ def getGitHash = { ->
     return stdout.toString().trim()
 }
 
+// defining this property in the file ~/.gradle/gradle.properties
```
pasin: How does this work on Jenkins?

bmeike: It should do exactly what it did before, except now it will produce an `x86_64` version as well.

Issue #1900, opened 2019-04-04
BlobInputStream does not return -1 when finished reading


RocketRider:
I am reading a blob from the DB and wait for it to finish reading but it never returns -1.

final InputStream inputStream = document.getBlob(reference).getContentStream();
final byte[] bytes = new byte[1024 * 128];
while ((read = inputStream.read(bytes)) != -1) {
    //Endless loop
}

After reading it returns zero

RocketRider:
Thank you for the fast fix

PR #1902, opened 2019-04-04
Fix reading from Blob stream not returning -1  when end of stream

pasin:
* c4stream_read return zero not -1 when end of stream.
* Fixed BlobInputStream to return -1 when the number of bytes is zero.

#1900


shared/src/main/java/com/couchbase/lite/Blob.java
```diff
@@ -535,7 +535,7 @@ public int read(byte[] b) throws IOException {
             try {
                 byte[] bytes = readStream.read(b.length);
```
bmeike: Would in be  better to `if (bytes.length <= 0) { return -1; }`, right here, before the copy?

bmeike:
I am cherry-picking this to master.

PR #1905, opened 2019-04-05
Patches for #1899 and #1900

bmeike:
Fix tests to expect -1 on BlobStream EOF
BlobStream.read return 0 when called with 0-length buffer
Update core submodule


bmeike:
@pasin , did you make these changes already, somewhere?

pasin:
I have patched this manually by using the fix from the master branch. We can close this PR.

PR #1922, opened 2019-06-19
Test for CBL-71: Clashing aliases

bmeike:
CBL-71 does not seem to affect the Java bindings.

### DIFF
--- a/lib/src/androidTest/java/com/couchbase/lite/QueryTest.java
+++ b/lib/src/androidTest/java/com/couchbase/lite/QueryTest.java
@@ -21,19 +21,14 @@ import android.os.Build;
 import android.os.Handler;
 import android.os.Looper;
 
-import com.couchbase.lite.internal.support.Log;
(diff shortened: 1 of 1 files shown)


bmeike:
The only significant change here is the new test at line 1255.  Everything else is just the formatter.

lib/src/androidTest/java/com/couchbase/lite/QueryTest.java
```diff
@@ -1232,6 +1252,16 @@ public void testQuantifiedOperatorVariableKeyPath() {
         //TODO
     }
 
+    // https://issues.couchbase.com/browse/CBL-71
+    @Test
+    public void testSelectWithAlias() throws Exception {
```
jayahariv: seems like this is redundant like iOS, as the same test is already been done in below test `testSelectAll`

bmeike:
Abandoning

PR #1923, opened 2019-06-25
Kludge away the JavaDoc errors

bmeike:
I've been meaning to do something about the javadoc errors during the build.  Apparently Jenkins is noticing them and failing the build, so it is now important.

Unfortunately the cure is pretty much as bad as the illness.


bmeike:
@borrrden has convinced me that this actually has no affect on the success of the build.  Abandoning.

PR #1924, opened 2019-07-02
Fix multiple retry tasks scheduled in replicator (CB-131)

pasin:
* Canceled the pending scheduled retry before scheduling a new one.
* Added some more verbose logs.


pasin:
CC: @bmeike

PR #1926, opened 2019-07-08
Clean up JNI and improve freeing native memory (CBL-130)

pasin:
1. Used only public API except for RefCounted
    1. Use FLSlice instead of slice.
    2. Use FLSliceResult instead of AllocSlice.
    3. Use FLEncoder instead of Encoder and fix some bugs in FLEncoder.
2. Instead of freeing FLSliceResult in finalize(), make sure the consumer codes call free() method. This will match the way FLSliceResult is used natively.
3. Instead of freeing FLEncoder in finalize(), make sure the customer codes call free() method. This will match the way FLSliceResult is used natively.
4. Ensured to reset the shared FLEncoder after using it.
5. Fixed bug in jstringSlice C++ class when the string is null. This seems to cause a lot of memory issue (bad_alloc) when changing to use FLSlice instead of slice.

### DIFF
--- a/lib/src/androidTest/java/com/couchbase/lite/internal/core/C4BaseTest.java
+++ b/lib/src/androidTest/java/com/couchbase/lite/internal/core/C4BaseTest.java
@@ -138,24 +138,29 @@ public class C4BaseTest {
 
     private byte[] createFleeceBody(Map<String, Object> body) throws LiteCoreException {
         FLEncoder enc = new FLEncoder();
-        enc.beginDict(body != null ? body.size() : 0);
-        for (String key : body.keySet()) {
-            enc.writeKey(key);
-            enc.writeValue(body.get(key));
-        }
-        enc.endDict();
-        return enc.finish();
+        try {
+            enc.beginDict(body != null ? body.size() : 0);
+            for (String key : body.keySet()) {
+                enc.writeKey(key);
+                enc.writeValue(body.get(key));
+            }
+            enc.endDict();
+            return enc.finish();
+        } finally { enc.free(); }
     }
 
     protected byte[] json2fleece(String json) throws LiteCoreException {
         boolean commit = false;
         db.beginTransaction();
+        FLSliceResult body = null;
         try {
-            byte[] bytes = db.encodeJSON(json5(json).getBytes()).getBuf();
+            body = db.encodeJSON(json5(json).getBytes());
+            byte[] bytes = body.getBuf();
             commit = true;
             return bytes;
         }
         finally {
+            if (body != null) body.free();;
             db.endTransaction(commit);
         }
     }
@@ -176,9 +181,7 @@ public class C4BaseTest {
             assertNotNull(output);
             Assert.assertArrayEquals(input, output);
         }
-        finally {
-            enc.free();
-        }
+        finally { enc.free(); }
     }
 
     @After
--- a/lib/src/androidTest/java/com/couchbase/lite/internal/core/C4BaseTest.java
+++ b/lib/src/androidTest/java/com/couchbase/lite/internal/core/C4BaseTest.java
@@ -160,7 +160,7 @@ public class C4BaseTest {
             return bytes;
         }
         finally {
-            if (body != null) body.free();;
+            if (body != null) body.free();
             db.endTransaction(commit);
         }
     }
--- a/lib/src/androidTest/java/com/couchbase/lite/internal/core/C4MutableFleeceTest.java
+++ b/lib/src/androidTest/java/com/couchbase/lite/internal/core/C4MutableFleeceTest.java
@@ -33,7 +33,7 @@ import org.junit.Test;
 
 import com.couchbase.lite.LiteCoreException;
 import com.couchbase.lite.internal.fleece.AllocSlice;
-import com.couchbase.lite.internal.fleece.Encoder;
+import com.couchbase.lite.internal.fleece.FLEncoder;
 import com.couchbase.lite.internal.fleece.FLValue;
 import com.couchbase.lite.internal.fleece.FleeceDict;
 import com.couchbase.lite.internal.fleece.FleeceDocument;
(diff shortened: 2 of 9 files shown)


pasin:
I fixed error reported by checkstyle and findbugs and did force-push.

lib/src/androidTest/java/com/couchbase/lite/internal/core/C4MutableFleeceTest.java
```diff
@@ -93,9 +104,14 @@ static void verifyDictIterator(Map<String, Object> dict) {
...
-        FLValue v = FLValue.fromData(fleece);
-        if (v == null) { return "INVALID_FLEECE"; }
-        return v.toJSON5();
+        try {
+            FLValue v = FLValue.fromData(fleece);
+            if (v == null) { return "INVALID_FLEECE"; }
+            return v.toJSON5();
+        }
+        finally {
+            if (fleece != null) fleece.free();
```
jayahariv: may be add a comment, describing about this release of the fleece object. So caller don't need to worry about it.

lib/src/androidTest/java/com/couchbase/lite/internal/core/C4QueryBaseTest.java
```diff
@@ -100,12 +100,14 @@ protected C4Query compile(String whereExpr, String sortExpr, boolean addOffsetLi
...
         C4QueryOptions opts = new C4QueryOptions();
-        C4QueryEnumerator e = query.run(opts, encodeParameters(params));
+        AllocSlice encodedParams = encodeParameters(params);
+        C4QueryEnumerator e = query.run(opts, encodedParams);
         assertNotNull(e);
         while (e.next()) {
             docIDs.add(e.getColumns().getValueAt(0).asString());
         }
-        e.free();
+        if (e != null) e.free();
```
jayahariv: assertNotNill(e) already doing, is _e != null_ necessary?

lib/src/androidTest/java/com/couchbase/lite/internal/core/C4QueryBaseTest.java
```diff
@@ -116,20 +118,25 @@ protected C4Query compile(String whereExpr, String sortExpr, boolean addOffsetLi
     protected List<List<List<Long>>> runFTS(Map<String, Object> params) throws LiteCoreException {
         List<List<List<Long>>> matches = new ArrayList<>();
         C4QueryOptions opts = new C4QueryOptions();
-        C4QueryEnumerator e = query.run(opts, encodeParameters(params));
+        AllocSlice encodedParams = encodeParameters(params);
+        C4QueryEnumerator e = query.run(opts,encodedParams);
```
jayahariv: space needed after the comma, in the parameters section. 
C4QueryEnumerator e = query.run(opts, encodedParams);

lib/src/androidTest/java/com/couchbase/lite/internal/core/C4QueryBaseTest.java
```diff
@@ -116,20 +118,25 @@ protected C4Query compile(String whereExpr, String sortExpr, boolean addOffsetLi
...
         assertNotNull(e);
         while (e.next()) {
             List<List<Long>> match = new ArrayList<>();
             for (int i = 0; i < e.getFullTextMatchCount(); i++) { match.add(e.getFullTextMatchs(i).toList()); }
             matches.add(match);
         }
-        e.free();
+        if (e != null) e.free();
```
jayahariv: assertNotNill(e) already doing, is _e != null_ necessary?

lib/src/androidTest/java/com/couchbase/lite/internal/core/C4BaseTest.java
```diff
@@ -138,24 +138,29 @@ private void deleteFile(String filename) {
...
-        enc.endDict();
-        return enc.finish();
+        try {
+            enc.beginDict(body != null ? body.size() : 0);
+            for (String key : body.keySet()) {
+                enc.writeKey(key);
+                enc.writeValue(body.get(key));
+            }
+            enc.endDict();
+            return enc.finish();
```
jayahariv: `L123 => kFleeceBody = createFleeceBody(body);`

Do we need to free this kFleeceBody in the tearDown?

lib/src/androidTest/java/com/couchbase/lite/internal/core/C4BaseTest.java
```diff
@@ -138,24 +138,29 @@ private void deleteFile(String filename) {
...
+        FLSliceResult body = null;
         try {
-            byte[] bytes = db.encodeJSON(json5(json).getBytes()).getBuf();
+            body = db.encodeJSON(json5(json).getBytes());
+            byte[] bytes = body.getBuf();
             commit = true;
             return bytes;
         }
         finally {
+            if (body != null) body.free();;
```
jayahariv: multiple semi colons;;

lib/src/androidTest/java/com/couchbase/lite/internal/core/C4MutableFleeceTest.java
```diff
@@ -50,27 +50,38 @@
...
     }
 
     static AllocSlice encode(Object obj) throws LiteCoreException {
-        Encoder enc = new Encoder();
-        enc.writeObject(obj);
-        return enc.finish();
+        FLEncoder enc = new FLEncoder();
+        try {
+            enc.writeValue(obj);
+            return enc.finish2();
```
jayahariv: what do you think about adding a comment saying something like, result should be freed after use.

lib/src/androidTest/java/com/couchbase/lite/internal/core/C4MutableFleeceTest.java
```diff
@@ -50,27 +50,38 @@
...
     }
 
     static AllocSlice encode(MRoot root) throws LiteCoreException {
-        Encoder enc = new Encoder();
-        root.encodeTo(enc);
-        return enc.finish();
+        FLEncoder enc = new FLEncoder();
+        try {
+            root.encodeTo(enc);
+            return enc.finish2();
```
jayahariv: same as above? do you think its useful?

lib/src/androidTest/java/com/couchbase/lite/internal/core/C4QueryBaseTest.java
```diff
@@ -116,20 +118,25 @@ protected C4Query compile(String whereExpr, String sortExpr, boolean addOffsetLi
...
     }
 
     private AllocSlice encodeParameters(Map<String, Object> params) throws LiteCoreException {
-        Encoder encoder = new Encoder();
-        encoder.write(params);
-        return encoder.finish();
+        FLEncoder encoder = new FLEncoder();
+        try {
+            encoder.write(params);
+            return encoder.finish2();
```
jayahariv: is it going to be repetitive, if we add a comment saying to free return type? I feels like, any methods, which needs particular care with the return type should have a note saying so. So that if anybody looks at the method signature can see it, and can avoid the mistake. but repetition.

jayahariv:
may be title can be, _clean up android tests to utilize the improved java source_.

lib/src/androidTest/java/com/couchbase/lite/internal/core/C4BaseTest.java
pasin: No. It's java byte[] array. It will get GC.

lib/src/androidTest/java/com/couchbase/lite/internal/core/C4BaseTest.java
pasin: Thanks.

lib/src/androidTest/java/com/couchbase/lite/internal/core/C4QueryBaseTest.java
pasin: I will rewrite this a bit.

lib/src/androidTest/java/com/couchbase/lite/internal/core/C4QueryBaseTest.java
pasin: Thanks.

lib/src/androidTest/java/com/couchbase/lite/internal/core/C4QueryBaseTest.java
pasin: I added the comment to AllocSlice and FLSliceResult itself.

PR #1933, opened 2019-10-09
2.6.2 release candidate

bmeike:
Add test for empty dict
Fix test for LiveQuery delay

Change-Id: If308ff977e387dd475af0130249c546178133e70

### DIFF
--- a/lib/src/androidTest/java/com/couchbase/lite/LiveQueryTest.java
+++ b/lib/src/androidTest/java/com/couchbase/lite/LiveQueryTest.java
@@ -84,17 +84,22 @@ public class LiveQueryTest extends BaseTest {
             .where(Expression.property("number1").greaterThanOrEqualTo(Expression.intValue(0)))
             .orderBy(Ordering.property("number1").ascending());
 
-        value = new AtomicInteger(0);
-        query.addChangeListener(executor, change -> value.incrementAndGet());
+        // There should be two callbacks:
+        //  - immediately on registration
+        //  - after LIVE_QUERY_UPDATE_INTERVAL_MS when the change gets noticed.
+        latch1 = new CountDownLatch(2);
+        ListenerToken token = query.addChangeListener(executor, change -> {
+            latch1.countDown();
+        });
 
         createDocNumbered(12);
         createDocNumbered(13);
         createDocNumbered(14);
+        createDocNumbered(15);
+        createDocNumbered(16);
 
-        // Ya, I know...
-        Thread.sleep(1000);
-
-        assertEquals(1, value.get());
+        latch1.await(5, TimeUnit.SECONDS);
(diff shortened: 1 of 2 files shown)


lib/src/androidTest/java/com/couchbase/lite/ResultTest.java
```diff
@@ -1082,4 +1082,23 @@ public void check(int n, Result r) throws Exception {
...
+        MutableDocument mDoc = new MutableDocument(doc1);
+        mDoc.setDictionary(key1, new MutableDictionary());
+        save(mDoc);
+
+        final Query query = QueryBuilder.select(SelectResult.property(key1))
+            .from(DataSource.database(db))
+            .where(Meta.id.equalTo(Expression.string(doc1)));
+
+        ResultSet results = query.execute();
+        for (Result result : results.allResults()) { result.toMap(); }
```
pasin: Should be assert if result.toMap() is empty?

bmeike: Wooo... good idea.  It is the `allResults()` that used to fail.  Still, that's a good test.  Will add.

lib/src/androidTest/java/com/couchbase/lite/ResultTest.java
```diff
@@ -1082,4 +1082,29 @@ public void check(int n, Result r) throws Exception {
...
+        mDoc.setDictionary(key1, new MutableDictionary());
+        save(mDoc);
+
+        final Query query = QueryBuilder.select(SelectResult.property(key1))
+            .from(DataSource.database(db))
+            .where(Meta.id.equalTo(Expression.string(doc1)));
+
+        ResultSet results = query.execute();
+        for (Result result : results.allResults()) {
+            assertNotNull(results);
```
pasin: I think this one should be outside for-loop.

bmeike: duh.  Of course.  ... and the check for `result` inside.
Thanks.

lib/src/androidTest/java/com/couchbase/lite/LiveQueryTest.java
```diff
@@ -84,17 +84,22 @@ public void testLiveQueryDelay() throws CouchbaseLiteException, InterruptedExcep
...
         createDocNumbered(13);
         createDocNumbered(14);
+        createDocNumbered(15);
+        createDocNumbered(16);
 
-        // Ya, I know...
-        Thread.sleep(1000);
-
-        assertEquals(1, value.get());
+        latch1.await(5, TimeUnit.SECONDS);
```
Sandychuang8: Do we need to assertTrue here?

PR #1937, opened 2019-10-24
Updated build.gradle to reflect CBL-Java restructure

pasin:


lib/build.gradle
```diff
@@ -139,26 +139,26 @@ android {
     externalNativeBuild {
         cmake {
             version "3.10.2"
-            path "${CBL_JAVA_DIR}/CMakeLists.txt"
+            path "${CBL_JAVA_DIR}/lib/CMakeLists.txt"
```
bmeike: I think it might be better to define CBL_JAVA_DIR to include 'lib'.  ... or, if necessary, to define a new var with the lib in it...

pasin: I will make the change.

PR #1945, opened 2020-01-14
CBL-632: Missing license in .zip file

bmeike:
Get correct license text into the .zip file

### DIFF
--- a/etc/jenkins/publish.sh
+++ b/etc/jenkins/publish.sh
@@ -9,7 +9,7 @@ MAVEN_URL="http://172.23.121.218/maven2/internalmaven"
 
 
 function usage() {
-    echo "Usage: $0 <release version> <build number> <artifacts path> <scratchspace path>"
+    echo "Usage: $0 <release version> <build number> <artifacts path> <workspace path>"
     exit 1
 }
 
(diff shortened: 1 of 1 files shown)


etc/jenkins/publish.sh
```diff
@@ -61,12 +61,12 @@ echo "======== Fetch library dependencies"
...
 
 echo "======== Create zip"
-ZIP_STAGING="${SCRATCH}/staging"
+ZIP_STAGING="${WORKSPACE}/staging"
 rm -rf "${ZIP_STAGING}"
 mkdir -p "${ZIP_STAGING}"
 pushd "${ZIP_STAGING}"
 cp "${ZIP_BUILD}/target/dependency/"*.jar . || exit 1
-curl "https://raw.githubusercontent.com/couchbase/build/master/license/couchbase-lite/LICENSE_${EDITION}.txt" -o LICENSE.TXT || exit 1
+cp "${WORKSPACE}/product-texts/mobile/couchbase-lite/license/LICENSE_${EDITION}.txt" .
```
pasin: Should the file be renamed to LICENSE.TXT as before?

bmeike: I don't know.  If you would prefer that, I can certainly make the change.  I assumed the "official" name was appropriate.

